PR #27, opened 2016-02-01
Allow path lookups in un-delimited splices

Nemo157:
Makes it easier when you want to reference structs/functions in other modules without importing them into the local context. E.g. just using `html!(w, $maud::PreEscaped("value"))` instead of having to `use maud::PreEscaped`.


lfairy:
Merged, cheers!

PR #28, opened 2016-02-01
Add a shorthand syntax to define element classes

Nemo157:
This is similar to how [slim](http://slim-lang.com) works, very useful when you're using a class heavy css framework. An example from the bootstrap docs

```
div class="input-group" {
  span class="input-group-addon" { "$" }
  input type="text" class="form-control" aria-label="Amount (to the nearest dollar)" { }
  span class="input-group-addon" { ".00" } 
}
```

becomes

```
.input-group {
  span.input-group-addon { "$" }
  input.form-control type="text" aria-label="Amount (to the nearest dollar)" { }
  span.input-group-addon { ".00" } 
}
```

### DIFF
--- a/maud_macros/src/parse.rs
+++ b/maud_macros/src/parse.rs
@@ -394,11 +394,7 @@ impl<'cx, 'i> Parser<'cx, 'i> {
         }
         if !classes.is_empty() {
             self.render.attribute_start("class");
-            let mut s = String::new();
-            for class in classes {
-              s = s + &*class + " ";
-            }
-            self.render.string(s.trim());
+            self.render.string(&classes.join(" "));
             self.render.attribute_end();
         }
         Ok(())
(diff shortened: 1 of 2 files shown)


lfairy:
Thanks for the work you've done so far! I'll do a proper look-through within the next week or so.

Nemo157:
Thanks for starting the project :smile:. If you agree with all these changes (or some subset) and want me to merge them together into a single pull request just let me know (there are trivial merge conflicts around the tests :frowning:).

lfairy:
No worries -- separate pull requests are easier to review, and the merge conflicts are easy to deal with anyway :)

As for the patch: I'm concerned about the ambiguity introduced by the shorthand div syntax. Since you can nest elements using just a space, it would be unclear whether `p .foo` means a `<p>` containing a `<div>` or a `<p>` with class `foo`. Similarly for splices: `$foo .bar` can be read two different ways.

The syntax which includes an explicit tag name looks good though. Can you modify your patch so that it only includes this variant?

Nemo157:
Sure.

lfairy:
Lgtm, thanks!

PR #30, opened 2016-02-01
Add support for a match keyword

Nemo157:


Nemo157:
I'm pretty sure that this will not cover every single case yet, and the error messages aren't the greatest (missing `,`'s normally cause a flow on error instead of highlighting the exact issue, at least the errors are generally around the right place though).

But, it does support more stuff than what is tested :smile:. Even using it inside attributes works fine (I'll add a test for that).

Nemo157:
Rebased onto latest master.

maud_macros/src/parse.rs
```diff
@@ -300,6 +312,100 @@ impl<'cx, 'i> Parser<'cx, 'i> {
...
+                    span: sp,
+                    render: self.render.fork(),
+                }.match_bodies());
+                break;
+            },
+            [ref tt, ..] => {
+                self.shift(1);
+                match_var.push(tt.clone());
+            },
+            [] => parse_error!(self, sp, "expected body for this #match"),
```
lfairy: This should say `@match`

maud_macros/src/parse.rs
```diff
@@ -300,6 +312,100 @@ impl<'cx, 'i> Parser<'cx, 'i> {
...
+            [ref tt @ fat_arrow!(), ..] => {
+                self.shift(1);
+                body.push(tt.clone());
+                break;
+            },
+            [ref tt, ..] => {
+                self.shift(1);
+                body.push(tt.clone());
+            },
+            _ => parse_error!(self, sp, "invalid #match pattern"),
```
lfairy: Same here

lfairy:
LGTM, thanks!

Issue #26, opened 2016-01-21
Syntax error within macro using `$(expr)`


Nemo157:
Just encountered a bug very similar to #23, just using `$(expr)` inside the html instead.

```rust
#[test]
fn foo() {
    macro_rules! to_string {
        ($($x:tt)*) => {{
            let mut s = String::new();
            html!(s, $($x)*).unwrap();
            s
        }}
    }

    let name = "Lyra";
    let s = to_string!(p { "Hi, " $(name) "!" });
    assert_eq!(s, "<p>Hi, Lyra!</p>");
}
```

gives

```
tests/tests.rs:324:47: 324:48 error: expected `*` or `+`
tests/tests.rs:324     let s = to_string!(p { "Hi, " $(name) "!" });
```

Nemo157:
As a test I tried changing it to `$(name)*` and got a `syntax error` from mauds parser. With a bit of work I managed to get it working with that syntax with both identifiers or expressions inside the `()` as a workaround [3d8763b](https://git.nemo157.com/forks/maud/commit/3d8763b53f71082114f2ec89111d79994849dcbf).

Nemo157:
The workaround doesn't seem to always work, not sure exactly what triggers the error but when I tried using it in some more complicated macros I would sometimes get `error: attempted to repeat an expression containing no syntax variables matched as repeating at this depth`. I'm guessing it's ok when it's at the terminal of the macro expansion, but if you have any repetition sequences at the same level it errors out.

I guess the best fix would be very similar to the fix for #21. Although with this it has the additional problem that it's in the usage of the macro, rather than the definition that you'd have to stop using `$`, i.e.

```rust
let s = to_string!(p { "Hi, " #splice (name) "!" });
```

in the above example. Maybe it would be worth dropping `$` entirely and finding an alternative shorthand interpolation syntax to better interact with `macro_rules`?

Nemo157:
I tried a basic replacement `$` → `#` and it seems to work ([602fdc](https://git.nemo157.com/forks/maud/commit/602fdc)).  The main issue is that it would make any of the special cases like `#call` basically be keywords, so if you had something like

```rust
let (animal, call) = ("cow", "moo");
let s = to_string!(p {
  "The " #animal " says " #call
}
```

it would fail to parse (or potentially parse with an incorrect output depending on the exact context). I think this is probably ok as it can be worked around with

```rust
let (animal, call) = ("cow", "moo");
let s = to_string!(p {
[...]

Issue #24, opened 2016-01-20
Can't have single use renderables


Nemo157:
Since `Render` takes itself by reference anything that implements it needs to support re-rendering multiple times. While working on supporting modified event streams in (second example at [`maud-pulldown-cmark`](https://git.nemo157.com/rust/maud-pulldown-cmark#example)) I found this limiting as `pulldown_cmark::Parser` isn't reusable, requiring me to wrap stuff up in closures to allow re-creating the parser when necessary. In most cases it's very unlikely that these structs would be kept around, so I was wondering if it were possible to have `fn render(self, w: &mut Write) -> fmt::Result` and provide implementations for references where needed.

Nemo157:
To see whether this would work I attempted to implement it (https://git.nemo157.com/forks/maud/commit/f1ecd07). Took a while to work out how to get it working, but finally realised that emulating the `Fn`, `FnOnce` hierarchy would work, so in my code I could just `impl RenderOnce` and consume the `Parser` while rendering. If you want this I could open a pull request (+ `RenderMut` maybe, might be useful for some other advanced renderable?).

Nemo157:
I threw together an implementation of `RenderMut` and some better tests for both [64565e8](https://git.nemo157.com/forks/maud/commit/64565e8).

lfairy:
Cheers for the report. I agree that having a way to render a value by-move would be useful. But I'm hesitant about making that the default. After all, `println!("{}", x)` borrows `x` rather than moving it.

As for `RenderMut`: do you have a use case for this trait? I don't want to add it in unless there's a real use for it.

Nemo157:
I can't think of any real use case for `RenderMut`. I just remembered that it will still default to borrowing a value if it implements `Render` because of the existence of `impl<'a, T: Render> RenderOnce for &'a T` and calling the method via `use ::maud::RenderOnce; $expr.render(&mut *$w)`. The compiler will auto-ref `$expr` till it finds the implementation of `RenderOnce` for `&$expr` then call that. I'll add a test that shows that in my changes for #26.

PR #31, opened 2016-02-02
Change symbology

### LINKED ISSUE #21: Maud and macro_rules don't play well together
This doesn't work:

```rust
macro_rules! greet {
    () => ({
        let mut result = String::new();
        let name = "Pinkie Pie";
        html!(result, p { "Hello, " $name "!" });
        result
    })
}
```

That's because `macro_rules!` tries to expand `$name` itself, when it should be handled by Maud instead.

We can probably work around that by adding a new `#splice` magic operator. So the code above would look like this:

```rust
// ...
html!(result, p { "Hello, " #splice name "!" });
// ...
```

This works because it doesn't have a `$` in it.

See rust-lang/rust#6994

Nemo157:
* Change `$` -> `^` and `#` -> `@`.
 * Add in support for render by move (with test showing it will still default to by-ref if needed).
 * Add in some more tests.
 * Remove symbol from `#else` (I guess this could break using an element called `else`, but who's gonna do that?)

Fixes #21, #24 and #26.

### DIFF
--- a/maud_macros/tests/tests.rs
+++ b/maud_macros/tests/tests.rs
@@ -205,13 +205,13 @@ mod control {
         for (number, &name) in (1..4).zip(["one", "two", "three"].iter()) {
             let mut s = String::new();
             html!(s, {
-                #if number == 1 {
+                @if number == 1 {
                     "one"
-                } #else if number == 2 {
+                } else if number == 2 {
                     "two"
-                } #else if number == 3 {
+                } else if number == 3 {
                     "three"
-                } #else {
+                } else {
                     "oh noes"
                 }
             }).unwrap();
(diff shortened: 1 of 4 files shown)


maud_macros/tests/tests.rs
```diff
@@ -327,6 +327,100 @@ fn splice_with_path() {
...
+        fn render(&self, w: &mut std::fmt::Write) -> std::fmt::Result {
+            w.write_str(self.0)
+        }
+    }
+
+    let mut s = String::new();
+    let r = R("pinkie ");
+    html!(s, ^r).unwrap();
+    html!(s, ^r).unwrap();
+    // R is not-Copyable so this shows that it will auto-ref splice arguments that implement Render.
```
Nemo157: Here's the test showing that the compiler auto-refs the spliced values.

lfairy:
:+1: for using trait auto-ref.

I'll have to disagree on removing the sigil from `else`. Though I admit it looks cleaner, as with the shorthand class syntax it makes the grammar ambiguous, and precludes us from adding XML support in the future. That's not worth removing one character over IMO.

The rest of the patch looks good, so I'll be happy to merge once this is addressed.

Nemo157:
Ah, yep, I did wonder why it would be needed, but XML support makes sense. I'll add it back.

PR #38, opened 2016-07-18
Add namespace support for element and attribute names

msifd:


maud_macros/src/parse.rs
```diff
@@ -547,18 +550,30 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
 
     /// Parses a HTML element or attribute name.
     fn name(&mut self) -> PResult<String> {
-        let mut s = match *self.input {
-            [ident!(_, name), ..] => {
+        macro_rules! ident_with_minuses {
```
lfairy: Does this need to be a macro? It looks like it can be written as a plain method.

lfairy:
Looks good, modulo the comment I made above.

lfairy:
Thanks!

Issue #49, opened 2016-10-09
Make benchmarks more realistic


lfairy:
The [existing benchmarks](https://github.com/lfairy/maud/tree/master/benchmarks/benches) are okay, but a bit simplistic. I'd love to have code that reflects the kind of use cases we see in the real world.

lfairy:
@TheNeikos would you be interested in this issue? I know you're working on a website that uses Maud; that project could be a good source to draw from.

TheNeikos:
Hm, right now I'm doing something similar to Rails (cause I like their way of organizing things):

- controller calls view with data
- view computes it with `html!` which usually includes several other components that call `html!`
- it is all passed to a layout that then outputs the whole thing

I wouldn't mind generalizing it a bit and making a benchmark for it, although I am not sure what one would benchmark in that case. Nested `html!`? (Because benchmarking my way of doing it would be out of scope of Maud think)

lfairy:
It's great that you're interested in this sort of thing!

My goals with these benchmarks are twofold:

1.  To give me something more realistic to benchmark against, so I don't optimize myself into a hole. We had that weird issue with inlining before, and I suspect that a larger benchmark would have been more helpful for that.

2.  To compare approaches with other template engines.

    Horrorshow, in particular, still uses a method based on writers/closures, rather than returning a `String` as Maud does today. We only see a difference in the two approaches when we start including templates in other templates (I think that's what you mean by nested `html!`?). One of the reasons I switched to `String`s was to make use cases like yours more convenient, so I'd like to see how it compares performance-wise.

In other words, benchmarking your way of doing it is *not* out of scope. I think copying the code from a reasonably-complex view, and calling it would some hard-coded data, would make a decent benchmark.

TheNeikos:
Alright, I will do that then. :+1:

TheNeikos:
Hm... Sorry that I didn't come back to you yet. I forgot to put it on my to-do list and it slipped my mind. It is there now though, so you should get the start of a PR tonight (GMT+2).

lfairy:
I don't think there's anything else to do for this issue.

PR #52, opened 2016-10-17
Add complicated Maud benchmark

TheNeikos:
As part of #49 I've extracted a simple but still somewhat more involved example.

If you would like anything else to be added or more examples I could do that

### DIFF
--- a/benchmarks/benches/complicated_maud.rs
+++ b/benchmarks/benches/complicated_maud.rs
@@ -9,7 +9,7 @@ use maud::Markup;
 #[derive(Debug)]
 struct Entry {
     name: &'static str,
-    score: usize,
+    score: u32,
 }
 
 mod btn {
--- a/benchmarks/benches/complicated_maud.rs
+++ b/benchmarks/benches/complicated_maud.rs
@@ -1,0 +1,102 @@
+#![feature(plugin, test)]
+#![plugin(maud_macros)]
+
+extern crate maud;
+extern crate test;
+
+use maud::Markup;
+
+#[derive(Debug)]
+struct Entry {
+    name: &'static str,
+    score: usize,
+}
+
+mod btn {
+    use maud::Render;
+
+    #[derive(Copy, Clone)]
+    pub enum RequestMethod {
+        Get, Post
+    }
+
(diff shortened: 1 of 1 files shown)


benchmarks/benches/complicated_maud.rs
```diff
@@ -0,0 +1,102 @@
...
+                            input.btn type="submit" value=(self.label) /
+                        }
+                    ).into_string())
+                }
+            }
+        }
+    }
+}
+
+fn layout<S: AsRef<str>>(title: S, inner: Markup) -> Markup {
```
lfairy: The use of `AsRef<str>` here is pretty clever! I wonder if it's a good idea to apply in general.

TheNeikos: I've taken a liking to start using all the `std::borrow` and `std::convert` traits wherever I can, to leverage the 'zero-cost' in Rust (or I hope so) this is my [favorite bit](https://github.com/TheNeikos/rustbreak/blob/master/src/lib.rs#L138) I wrote so far.

TheNeikos:
Using `Render` also removed the need for `PreEscaped` so there is that

benchmarks/benches/complicated_maud.rs
```diff
@@ -0,0 +1,102 @@
...
+
+extern crate maud;
+extern crate test;
+
+use maud::{Markup, PreEscaped};
+
+#[derive(Debug)]
+struct Entry {
+    name: &'static str,
+    score: u16,
```
TheNeikos: I've opted for an usize since space is not an issue (and it will be aligned anyway) I thought that letting the compiler decide on the right thing (tm) might be the best

lfairy: I was going off rust-lang/rust#22240 which recommends `u32` here.

It's not a performance/alignment issue, but a style one. From what I've seen, 32 bits is the "default" size, and what you should go for first.

TheNeikos: Alright! Sounds good, changed it

lfairy:
Looks great, thanks!

PR #61, opened 2016-11-13
Speed up `std::fmt::Write::write_str` for `maud::Escaper`.

utkarshkukreti:
I'm seeing roughly 25% speedup on both the normal and complicated
benchmarks present in `benchmarks/benches/maud.rs` and
`benchmarks/benches/complicated_maud.rs` respectively.

Before:

    test render_template ... bench:             461 ns/iter (+/- 140)
    test render_complicated_template ... bench: 1,562 ns/iter (+/- 366)

After:

    test render_template ... bench:             350 ns/iter (+/- 69)
    test render_complicated_template ... bench: 1,145 ns/iter (+/- 633)

(The error margins are pretty high in the above snippets but I've
ran these benchmarks several times and the difference is consistent.)

<hr>

What do you think @lfairy?

I also tried using `str::slice_unchecked` (there is no easy way to do unchecked slicing for slices yet; see https://github.com/rust-lang/rust/issues/35729) but there was no measurable improvement.

### DIFF
--- a/maud/src/lib.rs
+++ b/maud/src/lib.rs
@@ -176,14 +176,28 @@ impl<'a> Escaper<'a> {
 
 impl<'a> fmt::Write for Escaper<'a> {
     fn write_str(&mut self, s: &str) -> fmt::Result {
-        for b in s.bytes() {
-            match b {
-                b'&' => self.0.push_str("&amp;"),
-                b'<' => self.0.push_str("&lt;"),
-                b'>' => self.0.push_str("&gt;"),
-                b'"' => self.0.push_str("&quot;"),
-                _ => unsafe { self.0.as_mut_vec().push(b) },
+        let bytes = s.as_bytes();
+        let mut last = 0;
+        for (i, &b) in bytes.iter().enumerate() {
+            macro_rules! push {
+                ($str:expr) => {{
+                    unsafe {
+                        self.0.as_mut_vec().extend_from_slice(&bytes[last..i]);
+                    }
+                    self.0.push_str($str);
+                    last = i + 1;
+                }}
             }
+            match b {
+                b'&' => push!("&amp;"),
+                b'<' => push!("&lt;"),
+                b'>' => push!("&gt;"),
+                b'"' => push!("&quot;"),
+                _ => {}
+            }
+        }
+        unsafe {
+            self.0.as_mut_vec().extend_from_slice(&bytes[last..]);
         }
         Ok(())
     }


lfairy:
How does this compare when most of the characters need to be escaped? I know that some clever approaches fall over when given strings like `<<<<>>>><<<<>>>>` are involved.

TheNeikos:
@lfairy Maybe that could be added to the benchmarks?

utkarshkukreti:
@lfairy Good point. It's over 3.5x slower for escaping a long sequence of `<>` (13µs vs 3.5µs) and over 2x slower for escaping a long sequence of `<foo></foo>` (6.7µs vs 3µs). I guess the 2 benchmarks I compared in the commit message are the worst case for the current algorithm and best case for the new one. The question then is, what ratio of characters can we expect to require escaping in the real world and will this optimization be worth it for them?

lfairy:
@TheNeikos I'd love to see a escaping benchmark! Feel free to file a PR for that.

@utkarshkukreti The values being escaped are often provided by untrusted users (e.g. comments on a blog), so we need to assume the worst case. For this reason I'm wary of approaches which vary so much based on the input. There are ways to narrow the gap (using instructions like PCMPESTRI) but that will add much more complexity.

For these reasons, I'm going to close this PR. Thanks for your contribution regardless!

utkarshkukreti:
@lfairy Sounds reasonable. As for PCMPESTRI, I don't think it would be faster than the current implementation if the string contains only `<>&"` (but it would definitely be faster than the best case of this PR).

utkarshkukreti:
With a tiny change I'm able to get a huge improvement for `<><><>...`: 14µs (this PR) -> 8.5µs (with diff below).

```diff
--- a/maud/src/lib.rs
+++ b/maud/src/lib.rs
@@ -181,8 +181,10 @@ impl<'a> fmt::Write for Escaper<'a> {
         for (i, &b) in bytes.iter().enumerate() {
             macro_rules! push {
                 ($str:expr) => {{
-                    unsafe {
-                        self.0.as_mut_vec().extend_from_slice(&bytes[last..i]);
+                    if i > last {
+                        unsafe {
+                            self.0.as_mut_vec().extend_from_slice(&bytes[last..i]);
+                        }
                     }
                     self.0.push_str($str);
                     last = i + 1;
```

The benchmark:

```diff
--- a/benchmarks/benches/maud.rs
+++ b/benchmarks/benches/maud.rs
@@ -39,3 +39,17 @@ fn render_template(b: &mut test::Bencher) {
         }
     });
 }
+
+#[bench]
+fn render_escape(b: &mut test::Bencher) {
+    // let text = "<foo></foo>";
+    let text = "<><><><><>";
+
+    b.iter(|| {
+        html! {
+            @for _ in 0..100 {
+                (text)
+            }
+        }
+    });
+}
```

(Posting here for future reference... if I ever find a way to make this reasonably fast in the worst case.)

PR #67, opened 2016-12-05
Features/add meta tags to maud extras

ernestas-poskus:
1. Locking Cargo dependencies since master has [build error] it becomes tedious to work with maud_extras crate.

2. Adding Javascript element analogous to Css

3. Adding meta tags struct which generate <title> and meta description

I have Form struct ready, will add in a few days.

### DIFF
--- a/maud_extras/lib.rs
+++ b/maud_extras/lib.rs
@@ -108,3 +108,29 @@ impl<T: AsRef<str>> Render for Title<T> {
         }
     }
 }
+
+/// Generate <meta charset=""> element.
+///
+/// # Example
+///
+/// ```rust
+/// # #![feature(plugin)]
+/// # #![plugin(maud_macros)]
+/// # extern crate maud;
+/// # extern crate maud_extras;
+/// # use maud_extras::*;
+/// # fn main() {
+/// let markup = html! { (Charset("utf-8")) };
+/// assert_eq!(markup.into_string(),
+///            r#"<meta charset="utf-8">"#);
+/// # }
+/// ```
+pub struct Charset<T: AsRef<str>>(pub T);
+
+impl<T: AsRef<str>> Render for Charset<T> {
+    fn render(&self) -> Markup {
+        html! {
+            meta charset=(self.0.as_ref()) /
+        }
+    }
+}
--- a/maud_extras/lib.rs
+++ b/maud_extras/lib.rs
@@ -82,3 +82,29 @@ impl<T: AsRef<str>, U: AsRef<str>> Render for Meta<T, U> {
         }
     }
 }
+
+/// Generate <title> element.
+///
+/// # Example
+///
+/// ```rust
+/// # #![feature(plugin)]
+/// # #![plugin(maud_macros)]
+/// # extern crate maud;
+/// # extern crate maud_extras;
+/// # use maud_extras::*;
+/// # fn main() {
+/// let markup = html! { (Title("Maud")) };
+/// assert_eq!(markup.into_string(),
+///            r#"<title>Maud</title>"#);
+/// # }
+/// ```
+pub struct Title<T: AsRef<str>>(pub T);
+
+impl<T: AsRef<str>> Render for Title<T> {
+    fn render(&self) -> Markup {
+        html! {
+            title (self.0.as_ref())
+        }
+    }
+}
--- a/maud_extras/lib.rs
+++ b/maud_extras/lib.rs
@@ -56,3 +56,29 @@ impl<T: AsRef<str>> Render for Js<T> {
         }
     }
 }
+
+/// Generate <meta> elements.
+///
+/// # Example
+///
+/// ```rust
+/// # #![feature(plugin)]
+/// # #![plugin(maud_macros)]
+/// # extern crate maud;
+/// # extern crate maud_extras;
+/// # use maud_extras::*;
+/// # fn main() {
+/// let m = Meta("description", "test description");
+/// assert_eq!(html!{ (m) }.into_string(),
+///            r#"<meta name="description" content="test description">"#);
+/// # }
+/// ```
+pub struct Meta<T: AsRef<str>, U: AsRef<str>>(pub T, pub U);
+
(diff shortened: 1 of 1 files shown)


maud_extras/Cargo.toml
```diff
@@ -9,8 +9,8 @@ repository = "https://github.com/lfairy/maud_extras"
 description = "Extra bits and pieces for use in Maud templates."
 
 [dependencies]
-maud = { path = "../maud", features = ["iron"] }
-maud_macros = { path = "../maud_macros" }
+maud = { version = "0.14.0", features = ["iron"] }
```
TheNeikos: I agree that this is a bad idea, I wouldn't fix it like this, but fix it by freezing the rust version in the travis build

maud_extras/lib.rs
```diff
@@ -30,3 +30,31 @@ impl<T: AsRef<str>> Render for Css<T> {
...
+/// assert_eq!(markup.into_string(),
+///            r#"<script src="app.js" type="text/javascript"></script>"#);
+/// # }
+/// ```
+pub struct Javascript<T: AsRef<str>>(pub T);
+
+impl<T: AsRef<str>> Render for Javascript<T> {
+    fn render(&self) -> Markup {
+        html! {
+            script src=(self.0.as_ref()) type=("text/javascript") {}
```
TheNeikos: This might not work if the string has special characters as those need to be percent encoded rather than escaped in a HTML fashion. 
https://url.spec.whatwg.org/#url-code-points

maud_extras/meta.rs
```diff
@@ -0,0 +1,50 @@
...
+/// let markup = meta::Tags {
+///   title: "Meta title",
+///   description: "Meta description",
+/// };
+/// let mut got: String = String::new();
+/// markup.render_to(&mut got);
+/// assert_eq!(got, r#"<title>Meta title</title><meta property="description" name="description" content="Meta description">"#);
+/// # }
+/// ```
+pub struct Tags {
```
TheNeikos: This seems a bit restrictive, why not a Vector Tags? That way one could put whatever meta-attribute one needs.

maud_extras/meta.rs
```diff
@@ -0,0 +1,50 @@
...
+    pub title: &'static str,
+    pub description: &'static str,
+}
+
+impl Render for Tags {
+    fn render(&self) -> Markup {
+        html! {
+            title { (self.title) }
+            meta property=("description") name=("description") content=(self.description) /
```
lfairy: Literal strings shouldn't have parentheses around them.

    meta name="description" ...

maud_extras/Cargo.toml
ernestas-poskus: > lfairy

Reasoning is that it does not compile on my machine without locking versions.

ernestas-poskus: When switched to `nightly.20161205160138` it started working will remove commit

maud_extras/lib.rs
ernestas-poskus: :ok:

ernestas-poskus:
@lfairy updated not sure how to go with Meta 'static lifetime since if user will want dynamic meta tag it might be cumbersome to use, what do you think ?

TheNeikos:
@ernestas-poskus `'static` is anything that has a lifetime that is as long as the program (at least), so `String` for example is `'static`: https://is.gd/kJIEho

lfairy:
@ernestas-poskus This definition:

```rust
pub struct Meta<T: AsRef<str>, U: AsRef<str>>(pub T, pub U);
```

would be fine with me.

It's okay to have more type parameters here, since it will all be inferred for the user anyway.

ernestas-poskus:
updated

maud_extras/meta.rs
```diff
@@ -0,0 +1,80 @@
...
+///
+/// ```rust
+/// # #![feature(plugin)]
+/// # #![plugin(maud_macros)]
+/// # extern crate maud;
+/// # extern crate maud_extras;
+/// # use maud_extras::*;
+/// # fn main() {
+/// let m = meta::Meta("description", "test description");
+/// assert_eq!(html!{ (m) }.into_string(), r#"<meta name="description" content="test description">"#);
```
lfairy: This line should be wrapped (80 char limit for docs).

maud_extras/meta.rs
```diff
@@ -0,0 +1,80 @@
+extern crate maud;
```
lfairy: I think this should be merged into `lib.rs` -- there isn't enough here to justify a separate module.

maud_extras/meta.rs
```diff
@@ -0,0 +1,80 @@
...
+/// # #![plugin(maud_macros)]
+/// # extern crate maud;
+/// # extern crate maud_extras;
+/// # use maud_extras::*;
+/// # fn main() {
+/// let m = meta::Meta("description", "test description");
+/// assert_eq!(html!{ (m) }.into_string(), r#"<meta name="description" content="test description">"#);
+/// # }
+/// ```
+pub struct Meta<T: AsRef<str>, U: AsRef<str>>(pub T, pub U);
```
lfairy: 👍

maud_extras/meta.rs
```diff
@@ -0,0 +1,80 @@
...
+/// assert_eq!(markup.into_string(),
+///            r#"<title>Maud</title>"#);
+/// # }
+/// ```
+pub struct Title<T: AsRef<str>>(pub T);
+
+impl<T: AsRef<str>> Render for Title<T> {
+    fn render(&self) -> Markup {
+        html! {
+            title { (self.0.as_ref()) }
```
lfairy: Don't need braces here

```rust
    title (self.0.as_ref())
```

ernestas-poskus:
Updated everything according to last comments

lfairy:
LGTM, thanks!

ernestas-poskus:
Thank you for comments, makes me improve ! :+1: :octocat:

PR #74, opened 2017-02-11
Update Rocket dependency to version 0.2.0

Ferdi265:
I'm trying to use this library with Rocket, so I updated the dependency to version 0.2.0.

### DIFF
--- a/maud/Cargo.toml
+++ b/maud/Cargo.toml
@@ -14,4 +14,4 @@ Compile-time HTML templates.
 
 [dependencies]
 iron = { version = "0.5.1", optional = true }
-rocket = { version = "0.2.0", optional = true }
+rocket = { version = ">= 0.1.2, < 0.3", optional = true }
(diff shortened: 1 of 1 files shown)


lfairy:
Thanks, didn't realize Rocket 0.2 was released already.

Is there a way to declare support for both versions 0.1 and 0.2 simultaneously? That should ease the transition path for those still on Rocket 0.1.

TheNeikos:
I think `<=1.2` should work, per http://doc.crates.io/specifying-dependencies.html

lfairy:
I'd prefer `>= 0.1.2, < 0.3` instead since that keeps the lower bound

Ferdi265:
Changed to lfairy's suggestion.

lfairy:
Looks good, thanks!

lfairy:
Released as [0.16.1](https://crates.io/crates/maud/0.16.1).

Ferdi265:
Thanks, I'll start using this then 👍

PR #79, opened 2017-03-07
Add MetaProperty element

ernestas-poskus:
Adding convenience struct for generating meta tags that contain `property` instead of `name`.

Useful in a context of open graph tags http://ogp.me/.


lfairy:
Looks good to me, thanks!

Issue #80, opened 2017-03-14
Build broken since rustc nightly-2017-03-04


sanmai-NL:
`libsyntax` changes again ... I've been working on fixes myself for a while today, but the required set of changes appeared to grow quickly, and I wouldn't be able to fix them efficiently without being very familiar with `libsyntax` and `maud` internals. 

```
   Compiling maud_macros v0.16.2
error[E0425]: cannot find function `tts_to_parser` in module `parse`
   --> /home/sanmai/.cargo/bin/registry/src/github.com-1ecc6299db9ec823/maud_macros-0.16.2/src/parse.rs:126:33
    |
126 |         let mut parser = parse::tts_to_parser(self.render.cx.parse_sess, tts);
    |                                 ^^^^^^^^^^^^^ not found in `parse`

error[E0053]: method `fold_tt` has an incompatible type for trait
   --> /home/sanmai/.cargo/bin/registry/src/github.com-1ecc6299db9ec823/maud_macros-0.16.2/src/parse.rs:91:5
    |
91  |     fn fold_tt(&mut self, mut tt: &TokenTree) -> TokenTree {
    |     ^ expected enum `syntax::tokenstream::TokenTree`, found reference
    |
    = note: expected type `fn(&mut parse::FlattenNtFolder, syntax::tokenstream::TokenTree) -> syntax::tokenstream::TokenTree`
               found type `fn(&mut parse::FlattenNtFolder, &syntax::tokenstream::TokenTree) -> syntax::tokenstream::TokenTree`
```

bb010g:
Last version that compiles for me is `nightly-2017-03-04`.

lfairy:
Thanks for reporting! Sadly, we're in an annoying intermediate state right now. The latest changes pretty much force us to use the new procedural macro interface (rust-lang/rust#38356), but parts of this new system aren't finished yet.

My recommendation at this point is to pin to `nightly-2017-03-04` (thanks @bb010g) and wait for the dust to settle.

--------

As far as I know, the only thing blocking a port is this task:

> - [ ] Implement a minimal API for `proc_macro::TokenStream` as outlined in the RFC.

The current parser, which performs direct pattern matching on `&[TokenTree]`, will not work with the new interface. So we need a good replacement before switching over.

There are also these nice-to-haves:

> - [ ] Include a `TokenStream` quoter `proc_macro::quote!` behind the `proc_macro` feature gate.

This isn't a hard requirement, but the codegen would be nicer with this around.

[...]

PR #83, opened 2017-04-19
a temporary fix to work with Rocket

skariel:
see #80


lfairy:
Thanks for the PR!

I've fixed the underlying issue now, so we should no longer need this patch.

Issue #84, opened 2017-04-22
Remove `error!` macro


lfairy:
See https://github.com/lfairy/maud/blob/be88fc39d5d56370d3db020ef85f4f093ab45a60/maud_macros/src/parse.rs#L14-L19

This should really be a method on `Parser` instead.

Tasks:

- [ ] Add a `Parser::error()` method with a similar signature to the `error!()` macro
- [ ] Change all instances of `error!(...)` to `return self.error(...)`
- [ ] Remove the `error!()` macro

alisha17:
I am working on this issue!

PR #86, opened 2017-04-25
Remove 'error!' macro

alisha17:
This is a fix to issue #84 
Tasks done :
- Add a Parser::error() method with a similar signature to the error!() macro
- Change all instances of error!(...) to return self.error(...)
- Remove the error!() macro

### DIFF
--- a/maud_macros/src/parse.rs
+++ b/maud_macros/src/parse.rs
@@ -11,13 +11,6 @@ use syntax::tokenstream::{Delimited, TokenStream, TokenTree};
 use super::render::Renderer;
 use super::ParseResult;
 
-macro_rules! error {
-    ($cx:expr, $sp:expr, $msg:expr) => ({
-        $cx.span_err($sp, $msg);
-        return Err(());
-    })
-}
-
 macro_rules! at {
     () => (TokenTree::Token(_, Token::At))
 }
@@ -88,6 +81,12 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
         self.input = &self.input[n..];
     }
 
+    /// Returns error message attached to sp and stops compilation immediately.
+   fn error<T>(&self, span: Span, message: &'static str) -> ParseResult<T> {
+        self.cx.span_err(span, message);
+        Err(())
+   }
+
     /// Parses and renders multiple blocks of markup.
     fn markups(&mut self, render: &mut Renderer) -> ParseResult<()> {
         loop {
@@ -155,9 +154,9 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
             // ???
             _ => {
                 if let [ref tt, ..] = *self.input {
-                    error!(self.cx, tt.span(), "invalid syntax");
+                    return self.error(tt.span(), "invalid syntax");
                 } else {
-                    error!(self.cx, self.span, "unexpected end of block");
+                    return self.error(self.span, "unexpected end of block");
                 }
             },
         }
@@ -172,7 +171,7 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
             render.string(&s.as_str());
             Ok(())
         } else {
-            error!(self.cx, lit.span, "literal strings must be surrounded by quotes (\"like this\")")
+            return self.error(lit.span, "literal strings must be surrounded by quotes (\"like this\")")
         }
     }
 
@@ -193,7 +192,7 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
                 self.shift(1);
                 if_cond.push(tt.clone());
             },
-            [] => error!(self.cx, sp, "expected body for this @if"),
+            [] => return self.error(sp, "expected body for this @if"),
         }}
         // Parse the (optional) @else
         let else_body = match *self.input {
@@ -210,7 +209,7 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
                         self.shift(1);
                         Some(self.block(sp, d.stream(), render)?)
                     },
-                    _ => error!(self.cx, sp, "expected body for this @else"),
+                    _ => return self.error(sp, "expected body for this @else"),
                 }
             },
             _ => None,
@@ -235,7 +234,7 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
                 self.shift(1);
                 cond.push(tt.clone());
             },
-            [] => error!(self.cx, sp, "expected body for this @while"),
+            [] => return self.error(sp, "expected body for this @while"),
         }}
         render.emit_while(cond.into_iter().collect(), body);
         Ok(())
@@ -255,7 +254,7 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
                 self.shift(1);
                 pattern.push(tt.clone());
             },
-            _ => error!(self.cx, sp, "invalid @for"),
+            _ => return self.error(sp, "invalid @for"),
         }}
         let mut iterable = vec![];
         let body;
@@ -269,7 +268,7 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
                 self.shift(1);
                 iterable.push(tt.clone());
             },
-            _ => error!(self.cx, sp, "invalid @for"),
+            _ => return self.error(sp, "invalid @for"),
         }}
         render.emit_for(pattern.into_iter().collect(), iterable.into_iter().collect(), body);
         Ok(())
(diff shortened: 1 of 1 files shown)


maud_macros/src/parse.rs
```diff
@@ -155,9 +154,9 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
             // ???
             _ => {
                 if let [ref tt, ..] = *self.input {
-                    error!(self.cx, tt.span(), "invalid syntax");
+                    self.error(self.cx, tt.span(), "invalid syntax");
```
lfairy: Since `.error()` is a method, it has access to `self` -- so you don't need to pass in `self.cx` separately.

maud_macros/src/parse.rs
```diff
@@ -88,6 +81,12 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
         self.input = &self.input[n..];
     }
 
+    /// Returns error message attached to sp and stops compilation immediately.
+    fn error($cx:expr, $sp:expr, $msg:expr) {
```
lfairy: This is not valid syntax.

alisha17:
While running cargo test, I am getting this error:

error[E0554]: #[feature] may not be used on the stable release channel
 --> /home/alisha/maud/maud/src/lib.rs:1:1
  |
1 | #![feature(specialization)]
  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^

error: aborting due to previous error

error: Could not compile `maud`.

lfairy:
@alisha17 Maud requires the nightly version of Rust; it won't work on stable.

How did you install the Rust compiler? If you used rustup, then you should follow [these instructions](https://github.com/rust-lang-nursery/rustup.rs/blob/master/README.md#working-with-nightly-rust) to install the nightly toolchain.

alisha17:
I am trying to make a function like this:
fn error(cx: ??, sp: Span, msg: &'static str) {
        cx.span_err(sp, msg);
        return Err(());
    }
What is actually the type of cx?
My apologies if the doubt is too obvious, I am learning Rust !

lnicola:
I never used `maud`, but I suppose `cx` should have the type the macro (function now) was called with. Since the macro was called with `self.cx`, the type should probably be the same as the field's type: `&'cx ExtCtxt<'a>`.

alisha17:
What should be the return type of this error method ? 
Because in some functions where error method is used it expects return type as () and in some functions, ParseResult type.

lnicola:
Do you have an example in mind? I'm not sure it's possible

I think the issue is that you replaced `error!` with `self.error`. The macro expanded to code that `return`ed from the function. You should probably use `return self.error` or reorganize the code otherwise in places where `self.error` is not called at the end of the function. [This, for example](https://github.com/alisha17/maud/blob/7d980dc9866500f3029043e2474f4a7ac726e010/maud_macros/src/parse.rs#L155) used to return an `Err`, but now will return `Ok`, since calling `self.error` does not exit the function.

alisha17:
I'm sorry but can you please elaborate a bit? Please bear with me! :\

From what I understand, I should modify the functions to return Err type when calling self.error at the end of the function ?

lnicola:
I'll try to. First, take a look over the original source: https://github.com/lfairy/maud/blob/master/maud_macros/src/parse.rs#L103. You'll notice that the functions that use `error!` already return a `ParseResult<()>`. Your `error` method should return the same thing (like the version I proposed), since it's what the code uses to report errors.

Second, consider code such as this:

```rust
    fn markup(&mut self, render: &mut Renderer) -> ParseResult<()> {
        match *self.input {
            /* snip */
            _ => {
                if let [ref tt, ..] = *self.input {
                    error!(self.cx, tt.span(), "invalid syntax");
                } else {
                    error!(self.cx, self.span, "unexpected end of block");
                }
            },
        }
        Ok(())
    }
```

The `error!` macro gets expanded to:

```rust
    fn markup(&mut self, render: &mut Renderer) -> ParseResult<()> {
        match *self.input {
            /* snip */
            _ => { // HERE
                if let [ref tt, ..] = *self.input {
                    self.cx.span_err(tt.span(), "invalid syntax");
                    return Err(());
                } else {
                    self.cx.span_err(self.span, "unexpected end of block");
                    return Err(());
                }
            },
        }
        Ok(())
    }
```

It's clear that if the part marked with `HERE` executes, the function will immediately return `Err(())` (thus exit). It won't return the `Ok(())` on the last line, since it will no longer reach that line.

In your PR, however, you have:

```rust
    /* snip */
            _ => {
                if let [ref tt, ..] = *self.input {
                    self.error(self.cx, tt.span(), "invalid syntax");
                } else {
                    self.error(self.cx, self.span, "unexpected end of block");
                }
        },
    }
    Ok(())
}
```

Because there's no more `return` statement, nothing happens with the error returned by `self.error` and the function reaches the `Ok(())` line. To avoid this, you should write `return self.error(self.cx, tt.span(), "invalid syntax");` instead.

Am I making sense?

Then there's a syntax bit that introduces a small complication: in Rust you can omit the `return` at the end of the function (among other situations). In the example above, the final line says `Ok(())`, and not `return Ok(());` as you might expect. You can read about this here: https://doc.rust-lang.org/book/functions.html.

Because of this, code such as

```rust
fn blue() -> ParseResult<()> {
    /* snip */
    if something {
        self.error(self.cx, self.span, "something")
    } else {
        Ok(())
    }
}
````

will work without the explicit `return`, as the `if` is the last statement in the function.

**Comment by alisha17** [comment]
Yeah thanks a lott! I understood all of that.

But in this function 
` fn match_body(&mut self, sp: Span, render: &mut Renderer) -> ParseResult<Vec<TokenTree>> {
        let mut body = vec![];
        loop { match *self.input {
            [ref tt @ fat_arrow!(), ..] => {
                self.shift(1);
                body.push(tt.clone());
                break;
            },
            [ref tt, ..] => {
                self.shift(1);
                body.push(tt.clone());
            },
            _ => **return Err(())**,
        }}`

Again there will be return type problem since it expects type as Result<std::vec::Vec<syntax::tokenstream::TokenTree>, _> , whereas our return type is Result<(), _>.
Instead I just did return Err() but it won't display the custom message this way.
Should I create a separate error type for it and return inside Err()?

**Comment by lnicola** [comment]
Doesn't this work?

```rust
    fn match_body(&mut self, sp: Span, render: &mut Renderer) -> ParseResult<Vec<TokenTree>> {
        let mut body = vec![];
        loop { match *self.input {
            [ref tt @ fat_arrow!(), ..] => {
                self.shift(1);
                body.push(tt.clone());
                break;
            },
            [ref tt, ..] => {
                self.shift(1);
                body.push(tt.clone());
            },
            _ => return self.error(sp, "invalid @match pattern"),
        }}
```

lfairy:
I think you can just make the return type generic:

```rust
fn error<T>(&self, span: Span, message: &str) -> ParseResult<T> {
```

alisha17:
@lfairy and @lnicola  Thanks a lot! At every step, I learnt something new :)

maud_macros/src/parse.rs
```diff
@@ -88,6 +81,12 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
         self.input = &self.input[n..];
     }
 
+    /// Returns error message attached to sp and stops compilation immediately.
+   fn error<T>(&self, span: Span, message: &'static str) -> ParseResult<T> {
+    self.cx.span_err(span, message);
+    Err(())
+   }
```
lfairy: The indentation looks wrong here. Please make sure that all lines of code are indented by a multiple of 4 spaces.

maud_macros/src/parse.rs
```diff
@@ -155,9 +154,9 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
             // ???
             _ => {
                 if let [ref tt, ..] = *self.input {
-                    error!(self.cx, tt.span(), "invalid syntax");
+                   return self.error(tt.span(), "invalid syntax");
                 } else {
-                    error!(self.cx, self.span, "unexpected end of block");
+                   return self.error(self.span, "unexpected end of block");
```
lfairy: Same here.

maud_macros/src/parse.rs
```diff
@@ -563,3 +562,4 @@ impl<'cx, 'a, 'i> Parser<'cx, 'a, 'i> {
         Ok(render.into_stmts())
     }
 }
+
```
lfairy: Please remove this extra newline.

alisha17:
Done. Fixed the styling issues.

lfairy:
Thanks for sticking through this!

And a big thank you to @lnicola for helping out! I often don't have the time to respond myself, so I appreciate it when other people step up 🙂

PR #87, opened 2017-05-11
make rocket support public

flosse:


lfairy:
What's the motivation behind this change? `impl`s are global; you don't need to import them yourself.

flosse:
I got a

    rocket::response::Responder<'_>` is not implemented for `maud::PreEscaped<std::string::String>

this happed to because of *using the wrong rocket version* but *not* because of the private module :-\
So It's my fault. I'm sorry!

PR #94, opened 2017-07-14
Update to Rocket 0.3

bb010g:
Also fixes diff links in CHANGELOG.md. Releasing would require bumping `maud` and `maud_macros` to 0.17.0 due to `Responder::respond(self)` becoming `Responder::respond_to(self, &Request)`.


lfairy:
Looks good, thanks!

PR #96, opened 2017-07-30
Allow terminating void elements with semicolons (;)

lfairy:
When Rust and HTML syntax differ, Maud tends to side with Rust syntax. This can be seen with string literals, for example, where we use backslash escapes instead of HTML entities. Using `;` to terminate void
elements is consistent with this idea.

Moreover, this reduces confusion around the generated code. Maud does not insert an extra slash into the HTML output (as per the spec) but the syntax may imply otherwise.

This confusion may have been the cause of a [bug I found in the wild][1], where the code omitted the trailing slash on an `input` element. I suspect that this is because the author thought that a trailing slash in the Maud template would lead to one in the HTML output. Switching to semicolons would prevent this misconception.

[1]: https://github.com/anowell/quasar/pull/3

### DIFF
--- a/maud/tests/basic_syntax.rs
+++ b/maud/tests/basic_syntax.rs
@@ -54,14 +54,20 @@ fn nesting_elements() {
 
 #[test]
 fn empty_elements() {
-    let s = html!("pinkie" br/ "pie").into_string();
+    let s = html!("pinkie" br; "pie").into_string();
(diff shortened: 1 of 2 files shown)


lfairy:
Merged in 1785071f1b17daff032c8a21095b23b3db6f73f8

PR #101, opened 2017-08-25
Add `Clone` and `Copy` impl to `PreEscaped` via derive

### LINKED ISSUE #100: Impl Clone, PartialEq, Eq for Markup?
Should be possible since it's just a string. Are there any reasons not to derive those traits?

LukasKalbertodt:
Fixes #100


lfairy:
Thanks!

PR #104, opened 2017-10-06
Enforce nightly toolchain

sanmai-NL:
This file makes `rustup` select the nightly toolchain for this project, as it
requires. See https://github.com/rust-lang-nursery/rustup.rs/tree/533f09818fff23ba55a7c60474f3798d41bdd69d#the-toolchain-file

### DIFF
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,0 +1,1 @@
+nightly


lfairy:
That makes sense, thanks!

PR #106, opened 2017-11-01
Add conventional docs.rs badge

sanmai-NL:
I'll leave the decision to keep or remove the existing ‘API reference’ link to you.


bb010g:
Could you do this without messing up the alignment? Also, these should probably be squashed.

bb010g:
Before, the badges displayed in the header after the project name. They're now under the header.

lfairy:
Thanks! Putting them on a separate layout is fine by me.

Issue #90, opened 2017-06-25
Default implementation of render with asynchronous http


P-E-Meunier:
So, I'm trying to use maud for nest.pijul.com. So far I like it, but nest.pijul.com runs on exactly four threads, forked in the beginning, and is entirely nonblocking. More specifically:

- My async Postgres crate starts exactly *one* connection to the server *per thread*, i.e. currently four connections, independent from the number of clients or requests. It allocates two buffers per thread, once, and never reallocates or drops them.
- My async HTTP server allocates two buffers per connection, and reuses them if you use HTTP keep-alive.
- An SSH server allocating a constant number of buffers (like 8) per connection.

At least one buffer per connection is not really avoidable if you don't want to mix requests between clients, but the internals of a server should not need that.

So, in this context, Maud has the potential to allocate a single buffer per thread, and reuse it between clients, because it is never interrupted by IO. In a standard synchronous server, I agree with your comments on benchmarks (you would need some number of buffers *per client*), but async IO can allow you to have a constant memory use and constant number of allocations.

lfairy:
Great to hear that you like it :)

Note that the `Render` trait has a [`.render_to()`](https://lambda.xyz/maud/maud/trait.Render.html#method.render_to) method, which appends to an existing buffer. The code generator uses this method by default, so there's nothing to worry about on that front as long as `.render_to()` itself is written efficiently.

The issue, then, would be with the `html!` macro itself. As you've noticed, every call to `html!` allocates a new buffer, which is less than optimal when you control the event loop. To solve this problem, we would need to either (a) take an output buffer as a parameter, or (b) wrap the generated code in a closure that does the same.

In my experience, either of these solutions will make Maud less ergonomic to use. I also think that for most users, this trade-off is not worth it for them. So if we do end up adding a no-allocation option for Maud, I think it should be optional.

lfairy:
I wonder if we could use trait overloading to pick between the two approaches based on context. Something like this:

```rust
trait FromTemplate<F: FnOnce(&mut String)> {
    fn from_template(template: F) -> Self;
}

impl FromTemplate<F: FnOnce(&mut String)> for Markup {
    fn from_template(template: F) -> Markup {
        let mut buffer = String::new();
        template(buffer);
        PreEscaped(buffer)
    }
}

struct MarkupFn<F>(F);

impl FromTemplate<F: FnOnce(&mut String)> for MarkupFn<F> {
    // ...
}

[...]

PR #111, opened 2017-12-03
Add html_to! macro

abdullahibnnadjo:
Hi, this PR targets the feature requested in #90 

It's based on your current work in the `better-lints` branch.

It's now possible to write:
```rust
struct Css(&'static str);

impl Render for Css {
    fn render_to(&self, buffer: &mut String) {
        // With a mutable reference
        html_to! { buffer,
            link rel="stylesheet" type="text/css" href=(self.0);
        }
    }
}

fn main() {
    let mut s = String::with_capacity(100);

    // With the owned string in scope
    html_to!{ &mut s,
        (DOCTYPE)
        html {
            head (Css("style.css"))
            //…
        }
    };
    println!("{}", s);
}
```

All tests were rewritten to use html_test! instead of html! (a wrapper to test against html! and html_to! with its two forms)

Please tell me if you see anything I should change, like moving the new functions in their own file etc, I wasn't sure if it was the right place to put them.

### DIFF
--- a/maud/tests/html_test.rs
+++ b/maud/tests/html_test.rs
@@ -1,0 +1,36 @@
+#[cfg(test)]
+#[macro_export]
+macro_rules! html_test {
+    (bootstrap: { $($bootstrap:tt)* }, assert_eq: $a_eq:expr, markup: $($m:tt)+) => {
+        {
+            $($bootstrap)* {
+                let s = html! { $($m)+ }.into_string();
+                assert_eq!(s, $a_eq, "html!");
+            }
+        }
+
+        {
+            let mut s = String::new();
+            $($bootstrap)* {
+                html_to! { &mut s, $($m)+ };
+                assert_eq!(s, $a_eq, "html_to!, borrowing buffer");
+                s.clear();
+            }
+        }
+
+        {
+            let mut s = String::new();
+            fn to_borrowed_buffer(buffer: &mut String) {
+                $($bootstrap)* {
+                    html_to! { buffer, $($m)+ };
+                    assert_eq!(buffer, $a_eq, "html_to!, buffer already borrowed");
+                    buffer.clear();
+                }
+            }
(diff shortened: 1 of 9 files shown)


maud/tests/html_test.rs
```diff
@@ -0,0 +1,36 @@
+#[cfg(test)]
+#[macro_export]
+macro_rules! html_test {
```
lfairy: This code generation is pretty clever, but I think from a code coverage point of view we can get away with just a few specific tests for `html_to!`.

maud_macros/src/parse.rs
```diff
@@ -9,12 +9,46 @@ use proc_macro::{
...
+    Allocated,
+    Custom(BufferBorrow)
+}
+
+
+#[derive(Clone)]
+pub struct OutputBuffer {
+    ident: TokenTree,
+    buffer_type: BufferType
+}
```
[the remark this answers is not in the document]
abdullahibnnadjo: Thanks for the feedback! Yeah you're right, I'll try your to reimplement that following your recommandations. The first try wasn't successful, but I should be able to do that being more familiar with proc macros now

maud/tests/html_test.rs
abdullahibnnadjo: Sure, just keeping this for now until you're ok with the rest

sanmai-NL:
@abdullahibnnadjo: Could you tell us your status?

abdullahibnnadjo:
@sanmai-NL Yes, I don't have the time right now to do this, but I'll definitely work on that later

lfairy:
Heads up @abdullahibnnadjo -- I've merged the `better-lints` branch now, so please base on top of `master` instead :)

lambda-fairy:
Closing due to inactivity. Feel free to reopen if you're still interested (I think a rewrite would be cleaner at this point).

Issue #121, opened 2018-04-06
Build error on nightly (2018-04-05)


anxiousmodernman:
I'm doing a release build with `rocket` features. I get a big run of errors when compiling `maud_macros`. This is a pretty recent development, so I don't think it's due to any code change in maud itself.

![proc](https://user-images.githubusercontent.com/396004/38401710-32c31d7c-390c-11e8-8223-fcf64cbe6ad4.png)

_many other errors not shown_

[These are the most recent docs](https://doc.rust-lang.org/nightly/proc_macro/) I can find for the `proc_macro` crate. It looks like major changes to the API.

lfairy:
Thanks for reporting. I'll look into it over the coming weekend. (Unless you'd like to do the upgrade instead 😉)

anxiousmodernman:
I'll give it a shot and take some notes here.

It looks like a `TokenTree` is now an enum, not a struct, and so we express different token types as enum variants. Converting the struct syntax to the new enum shouldn't be so bad. 

Not sure about an equivalent to `TokenTreeIter` yet. I think they want us to deal with a `TokenStream` directly, [it has an `IntoIterator` impl](https://doc.rust-lang.org/nightly/proc_macro/struct.TokenStream.html#implementations).

anxiousmodernman:
Worked through `markup` and `markups`. So far I'm just letting complaints from my IDE (thanks RLS!) be my guide, until we are actually compiling under the new types. Then we can address correctness. Flagging this in the meantime: https://github.com/anxiousmodernman/maud/commit/4cd5451cec60797e0e2425235a355791442c8dfb#diff-285844f0c3c48a3b23fa1b5ec1140722R166

A `TokenTree::Group(x)` holds a `Group` that has a `stream() -> TokenStream` method, but the yielded stream is the tokens without the specified delimiter. I'll keep going and assume this is what we want, but I'm actually not sure.

anxiousmodernman:
A `TokenStream` can be built from an iterator of `TokenTree`. My idea is to change the signature of `match_arm` to support this. It yields a funky from_iter/into_iter dance, and look like the internals of `match_arm` will need some rework. 
![match_arms](https://user-images.githubusercontent.com/396004/38476826-abf1020e-3b64-11e8-8a24-c45c9efc844a.png)

anxiousmodernman:
This might be fixed. Interested parties can try testing with my fork:

```toml
# in Cargo.toml
[dependencies]
[...]

PR #122, opened 2018-04-09
Refactor to keep up with changes to proc_macro

anxiousmodernman:
Instead of a `kind` field containting a `TokenNode` variant, a
TokenTree is now an enum with variants of different types (Literal, Op,
Term, etc). Note that a TokenTree could be a sequence of TokenTrees if it is a
Group variant.

Other notes:

I'm unsure about the set_span call in Builder::emit_if, but I did not want to
throw away the passed in Span.

Parsing relies on destructuring references to the values associated with
TokenTree enum variants

Refs #121

### DIFF
--- a/maud_macros/src/build.rs
+++ b/maud_macros/src/build.rs
@@ -1,6 +1,5 @@
-use proc_macro::{Delimiter, Literal, Span, TokenNode, TokenStream, TokenTree};
+use proc_macro::{Delimiter, Group, Literal, Span, TokenStream, TokenTree};
 use proc_macro::quote;
-
 use maud_htmlescape::Escaper;
 
 pub struct Builder {
@@ -24,7 +23,7 @@ impl Builder {
         if !self.tail.is_empty() {
             let expr = {
                 let output_ident = self.output_ident.clone();
-                let string = TokenNode::Literal(Literal::string(&self.tail));
+                let string = TokenTree::Literal(Literal::string(&self.tail));
                 quote!($output_ident.push_str($string);)
             };
             self.stmts.push(expr);
@@ -35,7 +34,15 @@ impl Builder {
     /// Reifies the `Builder` into a raw list of statements.
     pub fn build(mut self) -> TokenStream {
         let Builder { stmts, .. } = { self.flush(); self };
-        stmts.into_iter().collect()
+
+        // use a Group here?
+        let mut tts: Vec<TokenTree> = Vec::new();
+        for s in stmts.into_iter() {
+            let i = s.into_iter();
+            tts.extend(i);
+        }
+
+        tts.into_iter().collect()
     }
 
     /// Pushes a statement, flushing the tail buffer in the process.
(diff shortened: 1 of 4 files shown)


maud_macros/src/lib.rs
```diff
@@ -27,21 +28,28 @@ pub fn html_debug(input: TokenStream) -> TokenStream {
...
+    // QUESTION: u64_suffixed or unsuffixed?
+    let size_hint = Literal::u64_unsuffixed(size_hint as u64);
     let stmts = match parse::parse(input, output_ident.clone()) {
         Ok(stmts) => stmts,
         Err(e) => panic!(e),
     };
     quote!({
         extern crate maud;
-        let mut $output_ident = String::with_capacity($size_hint as usize);
+        //let mut $output_ident = String::with_capacity($size_hint as usize);
```
anxiousmodernman: The compiler complains about `$size_hint` here

anxiousmodernman:
@lfairy This compiles, but I have introduced some parsing errors in the process. My hope is that we can work through these and keep the broad structure of the code intact.

lfairy:
I don't have time to look at this yet, but if you change the `html!` calls in the failing tests to `html_debug!` then the generated code will be logged to the console. From there you should be able to spot the problem.

maud_macros/src/build.rs
```diff
@@ -34,8 +36,17 @@ impl Builder {
...
-        stmts.into_iter().collect()
+
+        // use a Group here?
+        let mut tts: Vec<TokenTree> = Vec::new();
+        for s in stmts.into_iter() {
+            let i = s.into_iter();
+            tts.extend(i);
+        }
+
+        tts.into_iter().collect()
```
lfairy: I believe this change can be reverted once rust-lang/rust#49734 lands

maud_macros/src/parse.rs
```diff
@@ -309,24 +295,22 @@ impl Parser {
     }
 
     fn match_arms(&mut self) -> ParseResult<TokenStream> {
-        let mut arms = Vec::new();
+        let mut arms: Vec<TokenTree> = Vec::new();
```
lfairy: rust-lang/rust#49734 applies here as well

anxiousmodernman:
Life on the bleeding edge! :racehorse: :dash:

anxiousmodernman:
I think we're close. Tests pass for me locally, but on CI the hyper dependency doesn't compile. I'll clean up warnings and try to work it out.

lfairy:
Clippy failing on Hyper is a known issue (rust-lang#49643; there's a Clippy issue as well but I can't find it right now). Since it's not our fault I think we can ignore it for now 🙂

anxiousmodernman:
Alrighty. I'll squash these down.

anxiousmodernman:
https://github.com/rust-lang/rust/pull/49734 was merged 21 hours ago. I think we might need to wait another day before it's in nightly.
fromiterator

maud_macros/src/parse.rs
```diff
@@ -554,22 +553,21 @@ impl Parser {
...
     /// Rust code.
-    fn block(&mut self, body: TokenStream, span: Span) -> ParseResult<TokenTree> {
+    fn block(&mut self, body: TokenStream, _span: Span) -> ParseResult<TokenTree> {
         let mut builder = self.builder();
         self.with_input(body).markups(&mut builder)?;
-        Ok(TokenTree {
-            kind: TokenNode::Group(Delimiter::Brace, builder.build()),
-            span,
-        })
+        Ok(TokenTree::Group(Group::new(Delimiter::Brace, builder.build())))
```
lfairy: This code should call `.set_span()` on the generated `Group`

anxiousmodernman:
I've addressed the previous comments now that the iterator changes have landed.

PR #109, opened 2017-11-07
Remove old/irrelevant note

sanmai-NL:
There's [a website for Clippy lints](https://rust-lang-nursery.github.io/rust-clippy/master/index.html) now.
But I think this basic information is not the concern of this crate within the current more mature Rust ecosystem.

Issue #66, opened 2016-12-04
Add a lint plugin for detecting common anti-patterns


lambda-fairy:
We can lint these things:

* [x] Use `maud::DOCTYPE` instead of `PreEscaped("<!DOCTYPE html>")`

* [ ] Rather than writing `output.push_str(&html! { ... }.into_string())`, just implement the [`render`](https://docs.rs/maud/0.14.0/maud/trait.Render.html#method.render) method instead

* [ ] Use `format_args!` (or list the values individually) instead of `format!`

The following lints inspect the structure of the template itself. It's not clear yet how they can be implemented under the current design. We'll probably need to do some refactoring before these can work.

* [ ] Use `.foo` instead of `class="foo"`, and `#foo` instead of `id="foo"`

    + [ ] Use `.foo[bar]` instead of `class=@if bar { "foo" }`

* [ ] Warn against [`a target="..."` without `rel="noopener"`](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)

* [ ] Warn against dynamic data in CSS and JS contexts, e.g. `script { "var name = '" (user.name) "'" }`

* [ ] Warn about non-standard elements and attributes

Further suggestions welcome!

ernestas-poskus:
thus elements that contains empty tags like id="" or class="" could be removed to minimize html size

lfairy:
@ernestas-poskus that can be extended to other attributes like `style` and `rel` as well

ernestas-poskus:
> that can be extended to other attributes like style and rel as well

sure, all empty tags could be removed

lfairy:
> sure, all empty tags could be removed

Not all of them. `<script>` tags are almost always empty, for example. Empty tags can also be used as hooks for scripts and styling. And on an `<img>` tag, an empty `alt` attribute and a missing `alt` attribute [have different meanings][1].

[1]: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/img#Attributes

lfairy:
I have started improving the lint plugin on the [better-lints](https://github.com/lfairy/maud/compare/better-lints) branch. When I'm done, the proc macro will use compile-time "markers" to communicate the structure of the original template to the lint plugin.

lfairy:
#127 🎉

lambda-fairy:
[...]

PR #127, opened 2018-05-12
Change macro internals to use an explicit AST

lfairy:
See [comment](https://github.com/lfairy/maud/pull/127#issuecomment-389788303)

~~Partially addresses #66~~

~~TODO:~~

* [ ] ~~Document how to use the lint plugin~~
    * [ ] ~~Document the `maud_noopener` lint~~
* [x] ~~Update CONTRIBUTING.md (see #109)~~
* [ ] ~~Document how the marker system works?~~
* [ ] ~~Add Clippy-style UI tests~~


lfairy:
After some thought, I'm starting to think that this marker system isn't the best approach going forward.

We can get 90% of the way there using the [proc macro diagnostics API](https://doc.rust-lang.org/proc_macro/struct.Diagnostic.html), and the remaining 10% isn't really worth all this effort.

The reasons why I chose to implement these lints in an actual *lint plugin* are:

1. Better integration with built-in Rust lints, as they can be modified with `#[allow]` and `#[deny]` attributes etc.
2. Better type information. For example, we can check that the contents of a `<script>` element comprise only constant expressions.

(1) isn't a big deal, since we can extend the syntax of Maud to have the same functionality. The Rust interface isn't a perfect fit anyway -- for example, a hypothetical "unknown element" lint would need to take a list of allowed custom element names, but the Rust lint options don't provide more granularity than on/off. (Clippy has a [separate config file][1], which introduces its own issues.)

(2) is harder to argue against, but we can add specific markers to indicate a splice-within-a-`<script>`-tag without adding markers for everything else. This is much smaller in scope and easier to implement than what is proposed in this PR.

That said, some of these changes (e.g. the addition of an explicit AST) are nice to have regardless, so I'm happy to land those changes after reverting the lint-specific ones.

[1]: https://github.com/rust-lang-nursery/rust-clippy/blob/master/README.md#configuration

Issue #135, opened 2018-06-11
New feature: actix-web support


anxiousmodernman:
I'm intrigued by the actix-web framework. It should be possible to implement their `Responder` trait and include it behind feature flag `actix-web` in the manner of Rocket or Iron.

Some documentation on responses in actix-web:

* https://actix.rs/actix-web/actix_web/trait.Responder.html
* https://actix.rs/docs/response/

Note the associated types in `Responder`: `Error` and `Item`. These will require some additional implementations.

anxiousmodernman:
![maud_actix](https://user-images.githubusercontent.com/396004/41260831-ebee8628-6d8d-11e8-8f69-56512dbbbc1b.png)

PR #136, opened 2018-06-11
impl Responder for actix-web framework

anxiousmodernman:
It looks similar to the Rocket implementation, and functionally is also similar:
a syncronous response for PreEscaped<String>

Refs #135

### DIFF
--- a/maud/src/lib.rs
+++ b/maud/src/lib.rs
@@ -10,6 +10,7 @@
 
 #![doc(html_root_url = "https://docs.rs/maud/0.17.5")]
 
+#[cfg(feature = "actix-web")] extern crate actix_web;
 #[cfg(feature = "iron")] extern crate iron;
 #[cfg(feature = "rocket")] extern crate rocket;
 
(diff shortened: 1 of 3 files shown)


anxiousmodernman:
Ah the infamous "cannot link ring" more than once. My understanding is that we need separate test builds for separate features. Bummer. Hopefully that gets fixed this year.

maud/src/lib.rs
```diff
@@ -189,3 +190,17 @@ mod rocket_support {
...
+
+#[cfg(feature = "actix-web")]
+mod actix_support {
+    use PreEscaped;
+    use actix_web::{Responder, HttpResponse, HttpRequest, Error};
+
+    impl Responder for PreEscaped<String> {
+        type Item = HttpResponse;
+        type Error = Error;
+        fn respond_to<String>(self, _req: &HttpRequest<String>) -> Result<Self::Item, Self::Error> {
```
lfairy: This is better written as 

```rust
fn respond_to<S>(self, _req: &HttpRequest<S>) -> ...
```

to avoid confusion with the built-in `String` type

max-frai:
@lfairy Can you merge this into master now? Because I use same implementation locally :)

Issue #117, opened 2018-01-29
html attribute `for` (on label tag) doesn't parse


cschneid:
This works:

```
form action="/signup" method="post" {
    h1 "signup form"
    div {
        label "Email";
        input#email name="email";
    }
    div {
        label "Password";
        input#password type="password" name="password";
    }
    button type="submit";
}
```

But add `for` attribute to the labels and it stops compiling.

```
...
label "Email" for="email";
...
```

The error message emitted is:

```
error: proc macro panicked
  --> users_web/src/lib.rs:71:37
   |
71 |       MaudTemplate::new(public_layout(html! {
   |  _____________________________________^
72 | |         form action="/signup" method="post" {
73 | |             h1 "signup form"
74 | |             div {
...  |
83 | |         }
84 | |     }))
   | |_____^
   |
   = help: message: unexpected end of input
```

lfairy:
Thanks for reporting!

Does this work?

```rust
label for="email" "Email";
```

Element bodies should always go after attributes, as in HTML.

I've been thinking of making the braces mandatory, since I don't think the shorthand syntax has been pulling its weight. Maybe this issue is the kick I need to start working on that 😋

PR #137, opened 2018-06-16
Require braces around the body of an element

lfairy:
# Summary

Maud now requires braces around the body of an HTML element. For example, this code

```rust
html! {
    h1 a href="/" "Pinkie Pie's awesome website"
    div#main (contents)
}
```

must now be written as

```rust
html! {
    h1 {
        a href="/" { "Pinkie Pie's awesome website" }
    }
    div#main { (contents) }
}
```

This change was made to address usability issues, as well as improve visual harmony with surrounding Rust code.

# Rationale

Omitting the braces around an element body can reduce clutter. The feature is also easy to implement, and does not introduce ambiguity in the formal grammar.

However, over the last few months I have found many arguments for removing the feature:

* In #117, a user questioned why `label "Email" for="email"` didn't work. If we had required braces around the body, then it would be clear why `label { "Email" } for="email"` wouldn't parse.

* A void element (e.g. `input type="text";`) is always terminated by a semicolon. But an element with a single node in its body (e.g. `p "hello"`) does not require a semicolon. This is inconsistent with Rust syntax in general, which requires semicolons on statements unless they have return values or end with a braced block.

* The shorthand syntax also hides errors. If the user omits the trailing semicolon on a void element, then Maud will emit an extra end tag. For example, this code:

  ```rust
  html! {
      input name="cuteness" type="number"  // note missing `;`
      button "Update!"
  }
  ```

  results in the following HTML:

  ```html
  <input name="cuteness" type="number">
      <button>Update!</button>
  </input>
  ```

  It's possible for the library to detect and warn about these cases, but we should aim to avoid this situation in the first place.

# FAQ

## Do we need a deprecation period for this change?

I don't think so. Template engines tend to be direct dependencies, not transitive ones (like Serde), so it should be easy to pin an older version if you're not ready to deal with the breakage yet.

## I have another question. Should I ask it here?

Go for it! 😄

### DIFF
--- a/maud/tests/basic_syntax.rs
+++ b/maud/tests/basic_syntax.rs
@@ -51,12 +51,6 @@ fn simple_elements() {
 }
 
 #[test]
-fn nesting_elements() {
-    let s = html!(html body div p sup "butts").into_string();
-    assert_eq!(s, "<html><body><div><p><sup>butts</sup></p></div></body></html>");
-}
-
-#[test]
 fn empty_elements() {
     let s = html!("pinkie" br; "pie").into_string();
     assert_eq!(s, "pinkie<br>pie");
@@ -73,7 +67,7 @@ fn simple_attributes() {
     let s = html! {
         link rel="stylesheet" href="styles.css";
         section id="midriff" {
-            p class="hotpink" "Hello!"
+            p class="hotpink" { "Hello!" }
         }
     }.into_string();
     assert_eq!(s, concat!(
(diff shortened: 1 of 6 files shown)


anxiousmodernman:
Seems sensible. The unit tests read nicer, in my opinion.

Korvox:
This one bites. Went back to an older site to update it and got 150 errors of this, and fixing them all bloated one of my templates from 560 chars (without whitespace / control chars) to 644 chars, a 13% bloat.

I'm not sure how feasible it is, but I would love for it to be an option to disable redudnant braces. Its probably the Pythonista talking, and you are right that it is Rust and this isn't really the domain of terse beautiful syntax, but it did feel like magic to go from:

``` <button class="btn">Thing</button>```

to

```button.btn "Thing"```

Like sanity in a space where so little has existed for twenty years.

``` button.btn { "Thing" }```

Isn't worse than the original, but its for sure a lot less magical. 

In some ways it is even more redundant than regular html, since you have the braces and the quotations demarking the text body whereas with html you delineate the tags only, which kind of sucks.

Not really any point here, just going to miss how beautiful Maud was. It felt as close to no boilerplate markup as you could get. I'd definitely love some tunable to make the extra braces a warning or something.

Issue #129, opened 2018-05-13
Disallow elements that mention the same attribute twice


lfairy:
Duplicate attributes (e.g. `p title="Hello" title="world"`) are not allowed by HTML5, but Maud still accepts them. It would be nice if Maud emitted an error in this case.

This includes elements which have both a shorthand class/ID attribute and a longhand one (e.g. `p.foo class="bar"`), as they expand to duplicate attributes as well.

Note that this should be implemented as a hard error in `maud_macros`, not a lint, since there is no situation where duplicate attributes are useful.

PR #138, opened 2018-07-09
Disallow dupe attrs

TomPridham:
This is in response to #129. 

I'm not sure if this is the best way to do this or not. I also don't know if it should have a better error message or if there is a way to pinpoint where the error is. Right now it will just `panic` the whole html block, which could get confusing if it's a large block. 

This also doesn't currently compile when running `cargo test` because of the `panic` in the test. I thought that adding `#[should_panic]` would escape it. I'm not sure if I'm doing something wrong there or not. I could just remove that test, but I'm hoping it's something silly that you'll see.


TomPridham:
@lfairy hey, let me know if this looks good

maud_macros/src/parse.rs
```diff
@@ -558,6 +559,39 @@ impl Parser {
...
             }
         }
+
+        let mut attr_map: HashMap<String, Vec<Span>> = HashMap::new();
+        if let Some(class) = classes_static.first() {
+            attr_map.insert("class".to_owned(), vec![ast::span_tokens(class.clone())]);
+        }
+        if let Some((tokens, _)) = classes_toggled.first() {
+            attr_map.insert("class".to_owned(), vec![ast::span_tokens(tokens.clone())]);
+        }
```
TomPridham: im not sure if this is the best way or not. it has the side effect of showing the error on the first conditional class instead of whichever comes first. so . `p.test.hi[foo] class="whatever"` would error on `hi` and the actual `class` Attribute.

maud_macros/src/parse.rs
```diff
@@ -558,6 +559,39 @@ impl Parser {
...
+
+        for (name, spans) in attr_map {
+            if spans.len() > 1 {
+                let mut spans = spans.into_iter();
+                if let Some(first_span) = spans.next() {
+                    spans
+                        .fold(
+                            first_span.error(format!("duplicate attribute `{}` used here", name)),
+                            |acc, span| acc.span_error(span, format!("duplicate attribute `{}` used here", name))
```
TomPridham: I don't know if there is a more standard way to format this. i tried using `cargo fmt`, but that changed a bunch of files.

TomPridham:
no problem. thanks for all the pointers!

PR #145, opened 2018-08-09
Fixed benchmarks for liquid 0.15.0

lambdasqd:
Use of liquid::Object instead of liquid::Context and use of
Value::scalar instead of Value::Str and Value::Num. Changes to how the
parse() method is called.


lfairy:
Excellent, thank you!

PR #150, opened 2018-09-03
Implement streaming

### LINKED ISSUE #149: Support returning a Stream
Would be an interesting direction to explore. The main use-case would be streaming HTML to the browsers without blocking on database calls.

```rust
let name = futures::future::ok("<p>Maud</p>");

let stream = html_stream! {
  h1 { "Greetings, " }
  (name)
  p { "How are you?" } 
};
```

sorin-davidoi:
Quick-and-dirty implementation, contains lots of duplicate code. Will try to clean it up in the upcoming days.

Todo:
 - [x] Reduce code duplication
 - [ ] Handle escaping
 - [ ] Implement conversions (so you aren't required to pass futures)
 - [ ] More tests
 - [ ] Avoid boxing (if possible)
 - [x] Make it a feature (off by default)

Closes #149.

### DIFF
--- a/maud_macros/src/lib.rs
+++ b/maud_macros/src/lib.rs
@@ -81,7 +81,7 @@ fn expand_stream(input: TokenStream) -> TokenStream {
         extern crate maud;
         extern crate futures;
         let mut $output_ident: futures::stream::FuturesOrdered<
-            Box<futures::Future<Item = &str, Error = &str> + Send>
+            Box<futures::Future<Item = maud::Markup, Error = maud::Markup> + Send>
         > = futures::stream::FuturesOrdered::new();
         $stmts
         $output_ident
--- a/maud_macros/src/lib.rs
+++ b/maud_macros/src/lib.rs
@@ -36,11 +36,13 @@ pub fn html_debug(input: TokenStream) -> TokenStream {
     expr
 }
 
+#[cfg(feature = "streaming")]
 #[proc_macro]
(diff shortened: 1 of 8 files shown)


maud_macros/src/lib.rs
```diff
@@ -54,3 +68,22 @@ fn expand(input: TokenStream) -> TokenStream {
...
+    let markups = match parse::parse(input) {
+        Ok(markups) => markups,
+        Err(()) => Vec::new(),
+    };
+    let stmts = generate::generate_stream(markups, output_ident.clone());
+    quote!({
+        extern crate maud;
+        extern crate futures;
+        let mut $output_ident: futures::stream::FuturesOrdered<
+            Box<futures::Future<Item = &str, Error = &str> + Send>
```
sorin-davidoi: I think this should be `Box<futures::Future<Item = maud::Render, Error = maud::Render>>`, such that you can use anything that implements `maud::Render`. However I seem to be hitting the limits of my Rust skills as I can't see a way to make it compile. @lfairy do you have any ideas about how we might go about this?

lfairy: I believe that would have to be `Markup` or `Box<dyn maud::Render>` instead, as otherwise the consumer wouldn't know how much memory to allocate for each item in the stream?

sorin-davidoi:
Will close this for now. Might reopen at a later point.

PR #155, opened 2018-10-11
Added warning when elements have keyword names

### LINKED ISSUE #91: if and else are accepted instead of @if and @else
I just spent some time debugging one of the first maud codes I translated from mustache templates.
I had:

```
if is_editable {
   form …
}
```

Maud should have told me to use `@if` instead of `if`, with either an error or at least a warning.

ivanbakel:
Fixes #91.
To avoid erroneously preventing elements which are actually meant to be
called `if` from being used, this is only a warning, with a suggestion
of the template syntax.

### DIFF
--- a/maud_macros/src/parse.rs
+++ b/maud_macros/src/parse.rs
@@ -154,7 +154,7 @@ impl Parser {
                 match ident_string.as_str() {
                     "if" | "while" | "for" | "match" | "let" => {
                         ident.span()
-                            .warning(format!("found keyword `{}` - should this be a `@{}`?", ident_string, ident_string))
+                            .warning(format!("found keyword `{0}` - should this be a `@{0}`?", ident_string))
                             .emit();
                     }
                     _ => {}
--- a/maud_macros/src/parse.rs
+++ b/maud_macros/src/parse.rs
@@ -148,7 +148,17 @@ impl Parser {
                 }
             },
             // Element
-            TokenTree::Ident(_) => {
+            TokenTree::Ident(ident) => {
+                let ident_string = ident.to_string();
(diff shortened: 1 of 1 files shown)


ivanbakel:
I'd like to add a test for this, but I couldn't find anything on testing `Diagnostic` emission.

maud_macros/src/parse.rs
```diff
@@ -148,7 +148,17 @@ impl Parser {
...
             },
             // Element
-            TokenTree::Ident(_) => {
+            TokenTree::Ident(ident) => {
+                let ident_string = ident.to_string();
+                // Is this a keyword that's missing a '@'?
+                match ident_string.as_str() {
+                    "if" | "while" | "for" | "match" | "let" => {
+                        ident.span()
+                            .warning(format!("found keyword `{}` - should this be a `@{}`?", ident_string, ident_string))
```
lfairy: Is it possible to use the same argument twice? Something like

```rust
format!("found keyword `{0}` - should this be a `@{0}`?", ident_string)
```

lfairy:
> I'd like to add a test for this, but I couldn't find anything on testing `Diagnostic` emission.

Yeah, we'll need to use [compiletest-rs](https://github.com/laumann/compiletest-rs) to test that properly.

I think it's okay to skip the test for now.

maud_macros/src/parse.rs
ivanbakel: It was.

ivanbakel:
I had some time to get back to this, so I implemented a compiler test for the warning output. It's on a branch off this current one, if you don't want to merge it in as well.

PR #158, opened 2018-10-15
[WIP] Test `no_std` + `alloc` compat

sanmai-NL:
We use `maud` from a `no_std` + `alloc` crate. This does not work yet. My first step is to highlight the compatibility issue. Hopefully this leads to a simple fix. In the longer run, the best solution may be to switch `maud` over to being `no_std` + `alloc` itself. This is the natural solution, since `maud` is not dependent on OS specifics at all.

- [x] Add test for `no_std` + `alloc` compat.
- [x] Implement `no_std` + `alloc` compat for dependers.
- [ ] Adopt `no_std` + `alloc` within `maud`.


sanmai-NL:
Reopened to restart Travis CI.

sanmai-NL:
I’m exploring an upstream solution to the problem at https://github.com/dtolnay/quote/issues/80#issuecomment-429862280.

sanmai-NL:
@lfairy: Shall I squash this and leave the third item for another PR?

lfairy:
> Shall I squash this and leave the third item for another PR?

Yes, I think the `rust-toolchain` change should be discussed separately.

The purpose of the Travis build is to let us know of nightly breaking changes as they happen. That's why the build is set to run automatically every day, even when the code hasn't changed in the mean time. Locking the nightly to a specific version undermines that process.

sanmai-NL:
@lfairy: the `rust-toolchain` file specifies the default toolchain version, but CI builds are the exception rather than the rule. Would it be possible to have two continuous delivery pipelines:
- one that checks whether changes pass all tests given a known compatible toolchain version;
- one that checks whether the `maud` as recorded in the central repo (at one or more interesting versions) still compiles with the latest nightly toolchain.

sanmai-NL:
@lambda-fairy: I'm looking to process this stale PR. Is it better to close, or are some things of value?

lambda-fairy:
Thanks @sanmai-NL. The absolute paths issue was fixed a while ago, so I think I'll go ahead and close this.

If you're still interested, I'm still happy to take a PR for changing `std` to `alloc`.

PR #160, opened 2018-10-23
Set default content-type html for actix feature

infogulch:
Whenever I used `@for` of any kind in the returned Markup, Actix detected the content type as `application/octet-stream`, which prevented the browser from displaying it as html.

This PR explicitly sets content-type to "text/html" in the Actix feature handler, and fixes the above issue.  I don't see any reason why this would be incorrect, because all Markup instances are html.  One adjustment might be to explicitly set encoding to utf-8. Let me know if you'd prefer that.

### DIFF
--- a/maud/src/lib.rs
+++ b/maud/src/lib.rs
@@ -203,7 +203,9 @@ mod actix_support {
         type Item = HttpResponse;
         type Error = Error;
         fn respond_to<S>(self, _req: &HttpRequest<S>) -> Result<Self::Item, Self::Error> {
-            Ok(HttpResponse::Ok().body(self.0))
+            Ok(HttpResponse::Ok()
+               .content_type("text/html; charset=utf-8")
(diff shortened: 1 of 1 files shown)


lfairy:
That makes sense!

Yeah, I think explicitly setting the encoding to UTF-8 makes sense. It's what [Rocket does](https://api.rocket.rs/v0.4/rocket/http/struct.ContentType.html#associatedconstant.HTML) as well.

Leward:
Hello, is any update on the PR? Is there anything I can do to help?

lfairy:
@Leward As noted above, we should set the encoding to UTF-8 as well.

If you make a separate pull request with that change then I'll be happy to merge it.

infogulch:
Hi! Sorry it took me so long to get to this.  The code is now updated to include the encoding.

lfairy:
Thanks!

PR #162, opened 2018-12-11
bump rocket to 0.4

likewhatevs:
bump rocket version to 0.4

### DIFF
--- a/maud/Cargo.toml
+++ b/maud/Cargo.toml
@@ -16,7 +16,7 @@ categories = ["template-engine"]
 maud_htmlescape = { version = "0.17.0", path = "../maud_htmlescape" }
 maud_macros = { version = "0.19.0", path = "../maud_macros" }
 iron = { version = ">= 0.5.1, < 0.7.0", optional = true }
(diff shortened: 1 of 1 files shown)


maud/Cargo.toml
```diff
@@ -16,7 +16,7 @@ categories = ["template-engine"]
 maud_htmlescape = { version = "0.17.0", path = "../maud_htmlescape" }
 maud_macros = { version = "0.19.0", path = "../maud_macros" }
 iron = { version = ">= 0.5.1, < 0.7.0", optional = true }
-rocket = { version = "0.3", optional = true }
+rocket = { version = "0.4", optional = true }
```
lfairy: Can you change this to accept both 0.3 and 0.4? Just like with Iron and Actix.

```suggestion
rocket = { version = ">= 0.3, < 0.5", optional = true }
```

likewhatevs: Thanks for the tip, updated and good to know.

lfairy:
Thanks!

PR #166, opened 2019-03-24
Redesign the book

lfairy:
- The source files for the book are now maintained in this repository, not [maud-book](https://github.com/lfairy/maud-book).

- GitBook has been replaced with a custom static site generator that uses Maud itself.

- I also made some tweaks to the content. I've lost track of what's been changed, but I'm happy to run a diff if anyone wants it.

The redesigned book can be found at <https://maud.lambda.xyz/>.


lfairy:
Build is failing because of an issue in `compiletest_rs` (not related to this PR).

Issue #175, opened 2019-06-06
Compilation fails with the latest version of actix-web


Altenius:
actix-web 1.0.0 underwent a major rewrite, breaking maud since it still depends on 0.7.

```
...> cargo build
error: failed to select a version for `ring`.
    ... required by package `cookie v0.11.1`
    ... which is depended on by `actix-web v0.7.19`
    ... which is depended on by `maud v0.20.0`
    ... which is depended on by `... v0.1.0 (...)`
versions that meet the requirements `^0.13.0` are: 0.13.5

the package `ring` links to the native library `ring-asm`, but it conflicts with a previous package which links to `ring-asm` as well:
package `ring v0.14.6`
    ... which is depended on by `actix-http v0.2.3`
    ... which is depended on by `actix-web v1.0.0`
    ... which is depended on by `... v0.1.0 (...)`

failed to select a version for `ring` which could resolve this conflict
````

lfairy:
Thanks for reporting!

I'm busy with other things at the moment, but if you want to file a PR that bumps the dependency then I'll be happy to review it.

PR #176, opened 2019-06-10
Bumped actix-web version

Altenius:
The default features have also been disabled since they're unused by maud and caused issues when another dependency uses a different version of ring than actix-web.

### DIFF
--- a/maud/Cargo.toml
+++ b/maud/Cargo.toml
@@ -16,7 +16,7 @@ maud_htmlescape = { version = "0.17.0", path = "../maud_htmlescape" }
 maud_macros = { version = "0.20.0", path = "../maud_macros" }
 iron = { version = ">= 0.5.1, < 0.7.0", optional = true }
 rocket = { version = ">= 0.3, < 0.5", optional = true }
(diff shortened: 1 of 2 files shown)


Altenius:
Fixes #175

maud/Cargo.toml
```diff
@@ -16,7 +16,7 @@ maud_htmlescape = { version = "0.17.0", path = "../maud_htmlescape" }
 maud_macros = { version = "0.20.0", path = "../maud_macros" }
 iron = { version = ">= 0.5.1, < 0.7.0", optional = true }
 rocket = { version = ">= 0.3, < 0.5", optional = true }
-actix-web = { version = ">= 0.6.12, < 0.8.0", optional = true }
+actix-web = { version = "1.0.0", optional = true, default-features=false }
```
lfairy: Would be good to add spaces around the `=` for consistency

```suggestion
actix-web = { version = "1.0.0", optional = true, default-features = false }
```

Altenius:
Looks like I didn't have actix-web enabled when running the test. The Responder trait was changed to support futures; it should compile now.

lfairy:
Thanks!

PR #182, opened 2020-03-03
actix-web >= 2.0.0 support

czyba:
Cheers,

With the advent of async, actix-web started using futures in their APIs and traits, unfortunately resulting in breaking changes within the Responder trait which maud implements.

I have added support for the newest version of actix-web while trying to still support version 1, since most users have probably not yet migrated to the newest version.

In order to do so I added the trait `actix-web-2` feature and the `actix-web-2-dep` dependency in order to avoid name clashes. Unfortunately enabling the feature also depends on the `futures` crate, since it is used to implement the Responder trait. I also mentioned said support in the web-framework documentation. I have not added any tests, since there were no actix-web tests I could modify. I have verified that this change works within my own project though.

Note that `futures` dependency may not necessarily be required. It was used within actix-web itself to implement the Responder traits, but I could not trivially verify that it is strictly required to use that crate and not some other implementation.

P.S. Thanks for your awesome crate!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Change Log
 
 ## [Unreleased]
+- [Changed] Update Actix to 2.0.0. Actix 1.0.0 is no longer supported.
+  [#182](https://github.com/lambda-fairy/maud/pull/182)
 
 ## [0.21.0] - 2019-07-01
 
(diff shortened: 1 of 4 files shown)


lambda-fairy:
Thanks for your contribution!

How about we drop 1.x support and only support 2.x?

Users who are still using actix-web 1.x can pin to an older version of Maud, so I don't think we need to support both at once.

czyba:
Sure, I'll modify the PR accordingly!

czyba:
I removed the 1.0.0 support and added only the 2.0.0 support. I reverted the changes to the documentation as well, since they are no longer necessary.

I also took the liberty to modify the changelog accordingly.

lambda-fairy:
Merged, thank you!

PR #202, opened 2020-07-16
Add an off-by-default 'unstable' flag

### LINKED ISSUE #187: Add an off-by-default "unstable" feature
Add a Cargo feature named `unstable` to `maud`, which is off by default. This flag will be a no-op for now; it will be used in #188.

Add an "Unstable" build to the [Travis config](https://github.com/lambda-fairy/maud/blob/master/.travis.yml) which runs `cargo test` with the `unstable` feature turned on.

Change the "Benchmarks" build to enable the `unstable` feature as well (so that nightly-only optimizations are included).

# What about [`rustc_version`](https://crates.io/crates/rustc_version)?

It would be nice to use `rustc_version` to automatically toggle the feature flag based on whether the compiler is stable or nightly. However, as Maud doesn't build on stable yet, it would be impossible to test the flag if it worked this way. Using a manual flag instead lets us test the migration incrementally -- without having to execute it all at once. We can then slot in `rustc_version` when the migration is complete.

berwyn:
Closes #187

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,6 +9,12 @@ jobs:
         - |
           RUN_CLIPPY=true
           rustup component add clippy --toolchain=nightly || RUN_CLIPPY=false
+        # As of right now, these need to be tested individually becuase `--workspace`
+        # (which is implied by `workspace = true`) cannot be used with `--features`
+        # This _may_ be resolved with the `--package` flag, however this requires
+        # opting in to a `-Z` unstable feature as of time of commit.
+        #
+        # Follow https://github.com/rust-lang/cargo/issues/5364 for updates.
         - cargo test --manifest-path maud/Cargo.toml --features actix-web,iron,rocket
         - cargo test --manifest-path maud_htmlescape/Cargo.toml
         - cargo test --manifest-path maud_macros/Cargo.toml
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,9 @@ jobs:
         - |
           RUN_CLIPPY=true
           rustup component add clippy --toolchain=nightly || RUN_CLIPPY=false
-        - cargo test --all --features actix-web iron rocket
+        - cargo test --manifest-path maud/Cargo.toml --features actix-web,iron,rocket
+        - cargo test --manifest-path maud_htmlescape/Cargo.toml
+        - cargo test --manifest-path maud_macros/Cargo.toml
         - |
           if $RUN_CLIPPY; then
             CLIPPY_STATUS=0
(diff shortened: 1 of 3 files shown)


.travis.yml
```diff
@@ -18,9 +20,12 @@ jobs:
             done
             (exit $CLIPPY_STATUS)
           fi
+    - name: "Unstable"
+      script:
+        - cargo test --all --all-features
     - name: "Benchmarks"
       script:
-        - (cd benchmarks && cargo test --benches)
+        - (cd benchmarks && cargo test --benches --features unstable)
```
lambda-fairy, line 28: [Travis](https://travis-ci.org/github/lambda-fairy/maud/jobs/708836791) doesn't like this:

```
$ (cd benchmarks && cargo test --benches --features unstable)
    Updating crates.io index
error: Package `benchmarks v0.1.2 (/home/travis/build/lambda-fairy/maud/benchmarks)` does not have these features: `unstable`
The command "(cd benchmarks && cargo test --benches --features unstable)" exited with 101.
```

You might need to update `benchmarks/Cargo.toml` as well.

(There's also a [Clippy failure](https://travis-ci.org/github/lambda-fairy/maud/jobs/708836789), but that's not related to this PR so I'll fix that myself.)

.travis.yml
```diff
@@ -9,7 +9,9 @@ jobs:
         - |
           RUN_CLIPPY=true
           rustup component add clippy --toolchain=nightly || RUN_CLIPPY=false
-        - cargo test --all --all-features
+        - cargo test --manifest-path maud/Cargo.toml --features actix-web,iron,rocket
+        - cargo test --manifest-path maud_htmlescape/Cargo.toml
+        - cargo test --manifest-path maud_macros/Cargo.toml
```
lambda-fairy, line 14: It's good that you explained this in the commit message -- can you please include that info as a comment here as well?

If there's a relevant Cargo issue then that would be nice to mention too.

.travis.yml
lambda-fairy: (I've fixed the Clippy issue in #203.)

.travis.yml
berwyn: Can do! I'm not sure there's a relevant issue here, since it seems to be a very intentional choice but I'll see if there's anything relevant!

.travis.yml
berwyn: Cool, thanks! I've rebased the branch onto master, so that should resolve this!

.travis.yml
lambda-fairy: Thanks, looks great!

lambda-fairy:
Something strange is going on with Travis, but I'll go ahead and merge this in the mean time.

Thanks again @berwyn!

PR #207, opened 2020-08-30
Move to proc-macro-error and proc-macro2 and trybuild

### LINKED ISSUE #196: Add a shim for the unstable diagnostics API
Maud uses the proc macro diagnostics API for friendly error messages. This API is nightly-only, though, so we'll need to fall back to `compile_error!` or `panic!` on a stable compiler.

There is a [`proc-macro-error`](https://crates.io/crates/proc-macro-error) library that provides this functionality. We should investigate whether it meets our needs.

As discussed on #189, the "should be an `@if`" warning should be changed to an error as well.

CreepySkeleton:
Closes #196 

I settled on using macro syntax because I think it actually improves code readability (by avoiding extra `format!` noise). I remember @lambda-fairy expressed concerns of having to move from it to `proc_macro::Diagnostic` when it's stable. Let me allay you fears: one of the goals of `proc-macro-error` is to provide a shim that always chooses the best diagnostic implementation. Once compiler's built-in mechanism is stable, `proc-macro-error` (gosh, why did I choose so cumbersome name?) will just be delegating to it (as it does so currently on nightly), drawing excellent messages while requiring no changes on your side. Win-win.

The biggest change here is that `proc-macro-error` works best with `SpanRange` instead of `Span`, so I had to make extensive use of it, even though the changes are almost all mechanical. 

Besides moving to `proc-macro-error`, this PR also replaces all usages of `proc_macro::*` types with their `proc_macro2` counterparts. This doesn't really change anything because you already depend on it and extensively use (through `quote`).

Unresolved questions:

1. ~~You emit a warning in one place, but stable rustc has no way to emit custom warnings.  I can either turn it into a hard error or emit only on nightly (`proc-macro-error` has builtin support for this kind of warnings). Your call.~~

   Changing to a hard error as previously discussed.

2. https://github.com/lambda-fairy/maud/blob/a49552225928b8dcfef96c39764abbe7b5404dd5/maud_macros/src/parse.rs#L605-L612

   There's no way to add notes that point to a place other than the message itself on stable. I removed the notes.


CreepySkeleton:
Failure is due to the absent warning.

CreepySkeleton:
I'm trying to turn the warning into an error, but I'm getting _very weird_ failures with `comiletest-rs`. Would it be acceptable to just move to [`trubuild`](https://github.com/dtolnay/trybuild/)?

maud/tests/warnings/keyword_names_without_@.stderr
```diff
@@ -0,0 +1,7 @@
+error: found keyword `if`
+ --> $DIR/keyword_names_without_@.rs:9:9
```
lambda-fairy, line 2: Nit: is it good style to have an `@` symbol in a test filename?

Not sure what the convention is for trybuild test cases.

lambda-fairy (changes requested):
Overall looks great to me! I like how you've structured the work as separate commits – very clear.

Only concerns are with some `abort!` calls that should be `emit_error!` instead. I'm happy to merge once those are sorted out :smile:

lambda-fairy:
> There's no way to add notes that point to a place other than the message itself on stable. I removed the notes.

Is it possible to show that note only on nightly? (That might be a useful feature for proc-macro-error.)

maud_macros/src/parse.rs
```diff
@@ -89,28 +89,27 @@ impl Parser {
...
 
     /// Parses and renders a single block of markup.
-    fn markup(&mut self) -> ParseResult<ast::Markup> {
+    fn markup(&mut self) -> ast::Markup {
         let token = match self.peek() {
             Some(token) => token,
             None => {
-                Span::call_site().error("unexpected end of input").emit();
-                return Err(());
+                abort_call_site!("unexpected end of input");
```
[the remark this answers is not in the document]
CreepySkeleton: 👍

maud_macros/src/parse.rs
```diff
@@ -178,29 +176,25 @@ impl Parser {
...
-    fn literal(&mut self, lit: &Literal) -> ParseResult<ast::Markup> {
+    fn literal(&mut self, lit: &Literal) -> ast::Markup {
         let content = parse_str::<LitStr>(&lit.to_string())
             .map(|l| l.value())
-            .unwrap_or_else(|_| {
-                lit.span().error("expected string").emit();
-                String::new()  // Insert a dummy value
-            });
-        Ok(ast::Markup::Literal {
+            .unwrap_or_else(|_| abort!(lit, "expected string"));
```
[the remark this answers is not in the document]
CreepySkeleton: But then you don't have a literal to return if the error has triggered. I could return a dummy literal (like `"dummy"`) though; should I?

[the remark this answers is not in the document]
CreepySkeleton: Frankly, I on't see much point in this case. Missed literal should probably be a hard error, otherwise you're risking to spawn god knows how many non-existent errors while trying to recover.

maud/tests/warnings/keyword_names_without_@.stderr
CreepySkeleton: Well, I don't see `@` as a case of bad naming. If it was something obscure, sure, but the symbol is ASCII and can be typed with literally every keyboard. Anyway, if you don't happy, I could rename it to `keyword_names_without_at_sign` or something.

`trybuild` has no limitations for filenames AFAIK. You could even have an emoji in it if it was a wicked desire of yours...

maud_macros/src/parse.rs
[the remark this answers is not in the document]
lambda-fairy: The intention is to handle cases like

```rust
input type="number" min=0 max=100;
```

where the only problem is that the `0` and `100` need to be wrapped in quotes.

> A missed literal should probably be a hard error, otherwise you're risking to spawn a cascade of god knows how many non-existent errors while trying to recover.

That is true. But in this context, it's not a *missed* literal, just one of the wrong type. That said, I think the message can be improved... #208

> But then you don't have a literal to return if the error has triggered.

We can return `String::new()` like the original code.

maud/tests/warnings/keyword_names_without_@.stderr
lambda-fairy: Sounds good to me. My question was more whether this is *idiomatic*, rather than whether it works (the Travis build passes so I know it does). But it sounds like you've thought about this so I'm happy to resolve.

maud/tests/errors.rs
```diff
@@ -0,0 +1,9 @@
+use trybuild::TestCases;
+
+use std::path::PathBuf;
```
lambda-fairy, line 3:
```
warning: unused import: `std::path::PathBuf`
 --> maud/tests/errors.rs:3:5
  |
3 | use std::path::PathBuf;
  |     ^^^^^^^^^^^^^^^^^^
  |
  = note: `#[warn(unused_imports)]` on by default
```

---

I'm surprised that the `cargo clippy -- -D warnings` in `.travis.yml` didn't catch this. Can you please check if there's an easy way to lint tests as well? (If it's too hard then no worries – I can split out a new issue instead.)

CreepySkeleton:
```
cargo clippy --tests -- -D warnings
``` 

Be aware that `--tests` checks only the tests apparently, not the lib itself, so you're going to need two invocations:
```
cargo clippy -- -D warnings
cargo clippy --tests -- -D warnings
```

I had a bit of headache trying to figure it out four months ago. Maybe it has changed, don't know.

CreepySkeleton:
> Is it possible to show that note only on nightly? (That might be a useful feature for proc-macro-error.)

I'll repurpose `note = span => "message"`  syntax to be ignored on stable. Very few made use of it, and I think this kind of change would make their messages look better, not worse. Gimme a few days.

maud_macros/src/parse.rs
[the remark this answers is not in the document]
CreepySkeleton: > it's not a missed literal, just one of the wrong type

Right. Don't know why I thought it was about missing literals, let's just say that the voices in my head can be wrong sometimes.

maud/tests/errors.rs
lambda-fairy: Makes sense, thanks! I've filed this as #209.

lambda-fairy:
> I'll repurpose `note = span => "message"` syntax to be ignored on stable. Very few made use of it, and I think this kind of change would make their messages look better, not worse. Gimme a few days.

Thanks, that sounds like a good idea to me!

I've split this &ndash; and the CI issue &ndash; out into a separate ticket (#210) to avoid blocking this PR.

Thanks again for pushing this through!

PR #218, opened 2020-09-27
Update documentation to account for stable support

### LINKED ISSUE #217: Update documentation to account for stable support
Split from #213, as I don't want to confuse people who are still using the crates.io version.

lambda-fairy:
Closes #217


lambda-fairy:
Need to mention differences in behavior (specializations, diagnostics) if nightly is used.

PR #219, opened 2020-09-27
Enable stable support 🎉

### LINKED ISSUE #214: Toggle `unstable` flag automatically based on rustc version
The library should no longer expose a public `unstable` feature flag, and instead infer it using the [`rustc_version`](https://crates.io/crates/rustc_version) crate.

lambda-fairy:
Infer whether nightly features are available using `rustc_version`. If this fails, assume stable.

Closes #214

### DIFF
--- a/maud/build.rs
+++ b/maud/build.rs
@@ -1,0 +1,8 @@
+use rustc_version::{version_meta, Channel};
+
+fn main() {
+    match version_meta().channel {
+        Channel::Dev | Channel::Nightly => println!("cargo:rustc-cfg=unstable"),
+        Channel::Beta | Channel::Stable => {}
+    }
+}
--- a/maud/build.rs
+++ b/maud/build.rs
@@ -1,7 +1,7 @@
 use rustc_version::{version_meta, Channel};
 
 fn main() {
-    match version_meta().channel {
+    match version_meta().unwrap().channel {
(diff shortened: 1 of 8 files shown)


maud/build.rs
```diff
@@ -1,7 +1,7 @@
 use rustc_version::{version_meta, Channel};
 
 fn main() {
-    match version_meta().channel {
+    match version_meta().unwrap().channel {
```
CreepySkeleton, line 4: I don't recommend unwrapping in build scripts. I'm not familiar with `rustc_version` specifically, but this kind of "get compiler version" crates tend to return `None` or `Err` in case some part of the data could not be determined, [for example like the build date](https://gitlab.com/CreepySkeleton/proc-macro-error/-/issues/8). Although [it looks like this particular case is covered](https://docs.rs/rustc_version/0.2.3/src/rustc_version/lib.rs.html#161), simply ignoring `Err` is still a safer bet.

lambda-fairy: Thanks for chiming in. Yeah, given the low impact of defaulting to stable (slightly worse diagnostics, skipped specializations) I think that's the safer option – at least it's better than not working at all. And Cargo suppresses build script output [by design](https://github.com/rust-lang/cargo/issues/3777) so a warning won't work.

PR #227, opened 2020-10-07
Add a version string to the documentation

lambda-fairy:
This will be helpful for debugging once #226 lands.


lambda-fairy:
Screenshot

Not sure if I prefer it below the navigation or below the main content.

PR #228, opened 2020-10-07
Add actix-web v3 support

robjtede:
- adds actix-web v3 support
- retains v2 support
- thins deps by using futures-util

### DIFF
--- a/maud/examples/actix.rs
+++ b/maud/examples/actix.rs
@@ -1,0 +1,25 @@
+// do not use this line in your application
+use actix_web_dep as actix_web;
+
+use actix_web::{get, App, HttpServer, Result as AwResult};
+use maud::{html, Markup};
+use std::io;
+
+#[get("/")]
+async fn index() -> AwResult<Markup> {
+    Ok(html! {
+        html {
+            body {
+                h1 { "Hello World!" }
+            }
+        }
+    })
+}
+
+#[actix_web::main]
+async fn main() -> io::Result<()> {
+    HttpServer::new(|| App::new().service(index))
+        .bind(("127.0.0.1", 8080))?
+        .run()
+        .await
+}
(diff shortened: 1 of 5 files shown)


maud/src/lib.rs
```diff
@@ -7,13 +7,6 @@
 
 #![doc(html_root_url = "https://docs.rs/maud/0.22.0")]
```
robjtede, line 8: this could be `https://docs.rs/maud/0/maud/` so you don't have to update it as often

lambda-fairy: Yeah, I considered that but I'd rather keep it precise – I don't want someone's old build to have dead links just because the API changed.

A few respectable crates (e.g. [`syn`](https://docs.rs/syn/1.0.42/src/syn/lib.rs.html#253)) do the same thing so we're in good company.

robjtede: Ahh interesting. Wonder if upcoming intra-doc links will make this less necessary. Anyway I just hate having to remember this type of thing xD

maud/examples/actix.rs
```diff
@@ -0,0 +1,25 @@
...
+    })
+}
+
+#[actix_web::main]
+async fn main() -> io::Result<()> {
+    HttpServer::new(|| App::new().service(index))
+        .bind(("127.0.0.1", 8080))?
+        .run()
+        .await
+}
```
lambda-fairy, line 25: I'd love to integrate this better with the docs... either with some kind of `include` feature, or some ability to build code snippets like `rustdoc test`.

This doesn't block this PR; just an idle thought.

lambda-fairy: #230

lambda-fairy:
Thanks @robjtede!

PR #238, opened 2020-11-07
Make the `?` suffix for empty attributes optional

lambda-fairy:
# Summary

Maud no longer requires `?` after an empty attribute. For example, this code

```rust
html! {
    input type="checkbox" checked? disabled?[is_disabled];
}
```

can now be written as

```rust
html! {
    input type="checkbox" checked disabled[is_disabled];
}
```

The old syntax is still allowed, but soft-deprecated. It will be removed in some future release (TBD).

# Rationale

The `?` was necessary back when braces were optional, such that a bare identifier (e.g. `title`) could mean either an attribute or an element.

But with #137 making braces mandatory, there's no need for the `?` anymore – it is perfectly unambiguous without it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- [Changed] Don't require `?` suffix for empty attributes. The old syntax is kept for backward compatibility.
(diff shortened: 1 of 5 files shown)


lambda-fairy:
TODO:

- [ ] Update changelog
- [ ] Update docs

PR #242, opened 2020-11-29
Improve "elements and attributes" documentation

### LINKED ISSUE #239: Make curly brace requirement implicit where possible
Hey, thanks for making this library! I'm enjoying it so far :)
Just curious - I've spotted two place where perhaps curly braces can be implicitly assumed, rather than requiring the user to add clutter with them. These two situations are:

#### Elements without bodies

For example, when handling CSS links:
```
html! {
    html {
        head {
            link rel="stylesheet" href="/static/resident-smiler.css" {}
        }
    }
}
```
I feel like, in this case, the `{}` shouldn't be needed at the end of the `link` line? Maybe there's a shortcut I'm missing :)

#### When elements only have [a single line of] text as their child

For example:
```
html! {
    table {
        tr {
            td { "item 1 "}
            td { "item 2 "}
        }
        tr {
            td { "item 3 "}
            td { "item 4 "}
        }
    }
}
```
Versus:
```
html! {
    table {
        tr {
            td "item 1"
            td "item 2"
        }
        tr {
            td "item 3"
            td "item 4"
        }
    }
}
```

-----

I'm not sure that these are do-able due to parser limitations, but let me know!
Thanks again for the cool declarative HTML library :)

### LINKED ISSUE #241: Fails when numbers are provided in class names
[Bootstrap's grid system](https://getbootstrap.com/docs/4.0/layout/grid/#variable-width-content) involves classes with numbers, such as `col-sm-2`. This fails at compilation with the error: "expected string".

For example:
```
html! {
    .col-sm-2 {}
}
```

```
error: expected string
 --> file.rs:6:34
  |
n |     let markup = html! { .col-sm-2 {} };
  |                                  ^
```

lambda-fairy:
Closes #239
Closes #241

### DIFF
--- a/docs/content/elements-attributes.md
+++ b/docs/content/elements-attributes.md
@@ -1,14 +1,30 @@
 # Elements and attributes
 
-## Elements: `p`
+## Elements with contents: `p {}`
 
-Write an element using curly braces: `p { ... }`.
-
-Terminate a void element using a semicolon: `br;`. Note that the result will be rendered with HTML syntax – `<br>` not `<br />`.
+Write an element using curly braces:
 
 ```rust
 html! {
     h1 { "Poem" }
+    p {
+        strong { "Rock," }
+        " you are a rock."
+    }
+}
+```
+
+Before version 0.18, Maud allowed the curly braces to be omitted. This syntax was [removed][#137] and now causes an error instead.
+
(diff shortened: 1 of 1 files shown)


lambda-fairy:
cc @mitchhentges

mitchhentges (approved):
Looks good, thanks!

PR #276, opened 2021-06-11
Support for Rocket 0.5

cultpony:
The trait that Maud needs to implement is still the same in Release-Candidate 1 of Rocket 0.5, so simply raising the maximum version of rocket enables Maud to work.


cultpony:
I left this MR a bit too long in an open tab, saw there is another one, closing in favor of that.

PR #277, opened 2021-06-17
Warp framework support

bertdemiranda:
Maud can be used as such when using the Warp framework.
I added an example to the Framework support readme page for this.

### DIFF
--- a/docs/content/web-frameworks.md
+++ b/docs/content/web-frameworks.md
@@ -1,12 +1,13 @@
 # Web framework integration
 
 Maud includes support for these web frameworks:
-[Actix], [Iron], [Rocket], and [Rouille].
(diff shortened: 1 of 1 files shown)


lambda-fairy (commented):
Thanks for the PR! And sorry about the late response.

Looking at the code, I'm not sure if adding a Warp example is worthwhile.

At least with the Rouille example, we demonstrate the `Into<String>` impl which can save boilerplate.

But the Warp code doesn't show anything specific to Maud. In fact, you could replace the `hello_markup` function with any other library and the rest of the code would look the same.

Do other template libraries have examples like this?

bertdemiranda:
Yes, it would probably be sufficient to leave the example out and just mention that one has to apply a String::from.

PR #284, opened 2021-08-09
Add axum support

TheNeikos:
This PR adds support for [`axum`](https://docs.rs/axum/0.1.3/axum/)

### DIFF
--- a/maud/src/lib.rs
+++ b/maud/src/lib.rs
@@ -260,7 +260,7 @@ mod axum_support {
     use alloc::string::String;
     use axum::{
         body::Body,
-        http::{HeaderValue, Response, StatusCode},
+        http::{header, HeaderValue, Response, StatusCode},
         response::IntoResponse,
     };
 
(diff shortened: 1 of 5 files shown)


maud/src/lib.rs
```diff
@@ -253,3 +253,26 @@ mod tide_support {
...
+    use crate::PreEscaped;
+    use alloc::string::String;
+    use axum::{
+        body::Body,
+        http::{HeaderValue, Response, StatusCode},
+        response::IntoResponse,
+    };
+
+    impl IntoResponse for PreEscaped<String> {
+        fn into_response(self) -> Response<axum::body::Body> {
```
lambda-fairy, line 268:
```suggestion
        fn into_response(self) -> Response<Body> {
```

maud/src/lib.rs
```diff
@@ -253,3 +253,26 @@ mod tide_support {
...
+            let mut res = Response::new(Body::from(self.0));
+            *res.status_mut() = StatusCode::OK;
+            res.headers_mut().insert(
+                axum::http::header::CONTENT_TYPE,
```
lambda-fairy, line 272: Can we import the `header` module, and use `header::CONTENT_TYPE` here?

lambda-fairy (changes requested):
Thanks! Overall LGTM.

Can you please add an entry to the changelog as well?

lambda-fairy (approved):
Thanks!

PR #286, opened 2021-08-27
ensure tide default-features are disabled

maxcountryman:
This ensures that maud itself doesn't dictate what features an author might like to use.


lambda-fairy:
Thanks!

PR #287, opened 2021-08-30
report error for missing closing tag token

emi2k01:
image

As you can see, I used `abort_call_site` instead of adding a more appropriate span like the rustc one:

image

The reason is that I don't know how.


emi2k01:
I don't know how to make this a draft. The test failing is named "element-missing-body" but HTML tags are not required to have a body, I think.

Should I rename the test to `non-closed-element`?

maud_macros/src/parse.rs
```diff
@@ -553,6 +553,7 @@ impl Parser {
                     );
                 }
             },
+            None => abort_call_site!("expected ';' or '/', found end of macro"),
```
lambda-fairy, line 556: I'd rather only mention `;`, as the `/` syntax is soft-deprecated.

Also, Rust style is to use backticks to delimit code snippets.

```suggestion
            None => abort_call_site!("expected `;`, found end of macro"),
```

lambda-fairy (changes requested):
Thanks @emi2k01! Happy to merge once the comment is addressed (and the tests pass).

> Should I rename the test to `non-closed-element`?

Sure, do it!

emi2k01:
Btw, do you know if it's possible to create a diagnostic with whose span is just after the last token? Just like the second screenshot I added to the original comment.

lambda-fairy (approved):

lambda-fairy:
Thanks!

> Btw, do you know if it's possible to create a diagnostic with whose span is just after the last token?

There is `Span::after`, but it's unstable.

PR #289, opened 2021-09-04
Remove Iron support

lambda-fairy:
The Iron framework hasn't been updated in a long time. It still uses outdated versions of libraries like futures 0.1, hyper 0.12, url 1.7. We can clean up the repo by removing it. This will also slightly speed up CI.


lambda-fairy:
Benchmarked with `cargo clean; time cargo build --tests -j2`, run 3 times each

- Old: 3m36s
- New: 3m30s

lambda-fairy:
Two months and no-one's complained. Let's land this.

PR #297, opened 2021-09-19
Revert Cargo bug workaround

lambda-fairy:


lambda-fairy:
To be merged tomorrow, when rust-lang/cargo#9920 hits nightly.

PR #299, opened 2021-09-25
Restrict proc-macro2 version

GKnirps:
When I try to compile maud (rustc 1.54.0), I get the following error:

```
error[E0599]: no function or associated item named `from_str` found for struct `proc_macro::Literal` in the current scope
   --> /home/guido/.cargo/registry/src/github.com-1ecc6299db9ec823/proc-macro2-1.0.29/src/wrapper.rs:926:38
    |
926 |                 proc_macro::Literal::from_str(repr)
    |                                      ^^^^^^^^ function or associated item not found in `proc_macro::Literal`
```

Everything still works with proc-macro2-1.0.28, but versions older than 1.0.23 won't compile either. Of course, this version pinning is just a workaround, but it is a quick fix so maud can be built again.

- maud does not build with proc-macro2 < 1.0.23
- maud does not build with proc-macro2 >= 1.0.28

### DIFF
--- a/maud_macros/Cargo.toml
+++ b/maud_macros/Cargo.toml
@@ -14,7 +14,7 @@ edition = "2018"
 syn = "1.0.8"
 maud_htmlescape = { version = "0.17.0", path = "../maud_htmlescape" }
 quote = "1.0.7"
-proc-macro2 = "1.0.19"
+proc-macro2 = ">=1.0.23,<1.0.29"
 proc-macro-error = "1.0.0"
 
 [lib]


GKnirps:
Oh, and there seem to be other people who have problems with this change in proc_macro2 on rustc 1.54.0, see [this ticket](https://github.com/dtolnay/proc-macro2/issues/287).

GKnirps:
and I found out the problem does not occur anymore on rust 1.57.0-nightly. So this PR may not be required.

lambda-fairy:
Thanks for investigating.

Can this be closed? We don't have a formal MSRV, so the only requirement is that it works on the latest stable & nightly.

That said, I think bumping the lower bound is still helpful, so I'm happy to merge a PR with only that.

GKnirps:
Yes, I will close it and create a new PR that just bumps the dependency to the lowest version that works (i.e. 1.0.23).

PR #320, opened 2021-11-18
Remove blanket `Render` impl for `T: Display`

### LINKED ISSUE #271: Remove blanket `Render` impl for `T: Display`
While this is convenient, it's problematic because:

1. **Weird edge cases**. See #266. While that particular case could be solved with an `impl Render for &PreEscaped<T>`, I'd rather go for a more general solution.
2. **Performance**. The `std::fmt` machinery is pretty heavyweight, and building `Render` on top of it locks us out of making it faster. In fact, this is the reason why markup.rs beats Maud in benchmarks – they use the [`itoa` crate](https://crates.io/crates/itoa) to render integers.
3. **Semantics**. There's no particular reason why the text and HTML representations should be tied together.

As part of this change, we should:

- Add manual impls for integer types.
- Investigate impact on users like [miniserve](https://github.com/svenstaro/miniserve) (and maybe send them a migration PR)

lambda-fairy:
Closes #271

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+- Remove blanket `Render` impl for `T: Display`
+  [#320](https://github.com/lambda-fairy/maud/pull/320)
+
 ## [0.23.0] - 2021-11-10
 
 - Update to support axum 0.2
--- a/docs/Cargo.lock
+++ b/docs/Cargo.lock
@@ -326,6 +326,7 @@ checksum = "3e2e65a1a2e43cfcb47a895c4c8b10d1f4a61097f9f254f183aee60cad9c651d"
 name = "maud"
 version = "0.23.0"
 dependencies = [
+ "itoa",
  "maud_macros",
 ]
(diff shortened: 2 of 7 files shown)


lambda-fairy:
- [ ] Update changelog
- [ ] Update docs
- [ ] Add `itoa` optimization

lambda-fairy:
`miniserve` needs this patch:

```diff
diff --git a/src/renderer.rs b/src/renderer.rs
index ba2c4e6..72ff602 100644
--- a/src/renderer.rs
+++ b/src/renderer.rs
@@ -423,7 +423,7 @@ fn entry_row(
                         @if !raw {
                             @if let Some(size) = entry.size {
                                 span.mobile-info.size {
-                                    (size)
+                                    (size.to_string())
                                 }
                             }
                         }
@@ -432,7 +432,7 @@ fn entry_row(
             }
             td.size-cell {
                 @if let Some(size) = entry.size {
-                    (size)
+                    (size.to_string())
                 }
             }
             td.date-cell {
```

PR #322, opened 2021-11-26
Replace `PreEscaped` and `Markup` with `Html` type

### LINKED ISSUE #270: Redesign `PreEscaped` API
The current `PreEscaped` API has a few issues:

- The `PreEscaped` / `Markup` naming was lifted from `blaze-markup`, which was intended to support both HTML and XML. But Maud was always HTML-only, and the upcoming context-aware escaping effort will deepen this specialization.
    - Let's rename it to just `Html`.
- `PreEscaped` wraps any `T: AsRef<str>`, but I've only seen it used with `String` and `&'static str`.
    - Let's make it wrap `Cow<'static, str>` instead.
- The `PreEscaped` constructor makes it too easy to treat any arbitrary string as HTML. Modern APIs like the [Trusted Types proposal](https://developer.mozilla.org/en-US/docs/Web/API/TrustedHTML) force the user to do some sanitizing/escaping first, or at least acknowledge the security risk.
    - Let's remove the public constructor, and replace it with...
      ```rust
      impl Html {
          pub sanitize(value: &str) -> Self;
          pub from_trusted(value: impl Into<Cow<'static, str>>) -> Self;
      }
      ```
      Notice how the safe (`sanitize`) option is shorter!

lambda-fairy:
Closes #270


alexkazik:
I've tried this (together with rocket 0.5 support) and it works great.

vcavallo:
I'm really hoping this gets released because I'd like to use maud with HTMX.

PR #325, opened 2021-12-05
Update to axum-core 0.1

davidpdrsn:
Together with the new 0.4 release of `axum` we made a new crate called
[`axum-core`]. `axum-core` has a small API and is less likely to receive
breaking changes. Therefore its recommended for libraries to depend on
`axum-core` instead of `axum`.

So this updates `maud` to use `axum-core`.

Note this is a breaking change since `axum-core` requires `axum` 0.4.

[`axum-core`]: https://crates.io/crates/axum-core


lambda-fairy:
Thanks!

PR #328, opened 2022-01-07
Use explicitly-named args in `impl_render_with_display`

malyn:
Implicitly-named format args were introduced in 69649421686a16206d55eb31634f2b77f8aa504e as part of the work on #320, but it does not appear that feature will land until Rust 1.58. Explicitly naming the args ensures that the Maud works on non-Nightly builds of Rust.


lambda-fairy:
Thanks @malyn !

I tweaked the formatting to match that of the [std docs](https://doc.rust-lang.org/std/macro.format.html#examples). But otherwise LGTM.

PR #333, opened 2022-03-12
Add Poem support

kitified:
Adds support for [Poem](https://docs.rs/poem/latest/poem/) `IntoResponse`.

I didn't add any tests: I didn't see any for `axum`, and this is a similar type-level change.
It also looks like we don't have a great way to test stuff like this in general in the repo right now.

I did test with a local project, attached:
[maud_example.zip](https://github.com/lambda-fairy/maud/files/8238367/maud_example.zip)

### DIFF
--- a/maud/Cargo.toml
+++ b/maud/Cargo.toml
@@ -26,6 +26,7 @@ futures-util = { version = "0.3.0", optional = true, default-features = false }
 actix-web-dep = { package = "actix-web", version = "4", optional = true, default-features = false }
 tide = { version = "0.16.0", optional = true, default-features = false }
 axum-core = { version = "0.1", optional = true }
+poem = { version = "~1", optional = true }
 http = { version = "0.2", optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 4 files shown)


lambda-fairy (changes requested):
Thanks for the PR! And sorry for the delay :sweat_smile: 

> I didn't add any tests: I didn't see any for `axum`, and this is a similar type-level change.

Please add a `poem` dependency to the [`doctest` crate](https://github.com/lambda-fairy/maud/blob/main/doctest/Cargo.toml). This will check that the example you added to the docs can be built. (And also fix the CI failure.)

maud/Cargo.toml
```diff
@@ -26,6 +26,7 @@ futures-util = { version = "0.3.0", optional = true, default-features = false }
 actix-web-dep = { package = "actix-web", version = "4", optional = true, default-features = false }
 tide = { version = "0.16.0", optional = true, default-features = false }
 axum-core = { version = "0.1", optional = true }
+poem = { version = "~1", optional = true }
```
lambda-fairy, line 29: Any particular reason for the `~` dependency? I'd rather avoid these uncommon constructs unless there's a specific reason to use them.

```suggestion
poem = { version = "1.3", optional = true }
```

kitified:
Actually I've stopped using poem 😶 closing this.

giodamelio:
I am a current user of poem, and this would be nice so I don't need the wrapper struct.

If I resubmit this as a new PR, would you be open to it @lambda-fairy?

PR #338, opened 2022-04-10
Document that the `doctest` crate might need updating

lambda-fairy:
Also rename the "Examples" build to "Doctest" to clarify what it does.


lambda-fairy:
cc @CBenoit @kitified

Issue #62, opened 2016-11-14
How about XML template support? Just for a ask


mindcat:
Well, I take a look about maud, so I thought could we reuse syntax for XML template support.
> XML is a markup language much like HTML

lfairy:
I'll be keen for a corresponding `xml!` macro. AFAIK the only difference is that empty elements end with `/>` instead of just `>`.

Not sure how we'd prevent mixing HTML and XML code together though. Since the languages are slightly different, we have to enforce that they are separate somehow.

lfairy:
Marking as "hard", since there are some unanswered design questions (though I believe the actual coding will be easy once it's figure out).

gauteh:
Would be really interested in this as well!

th-we:
Did a quick test with XML generation and maud seems already usable for many use cases. What is it that should be supported for XML?

* Respecting Namespaces. This is invalid XML because there are two attributes with the same name and the same namspace URI:
  ```
  <foo xmlns:a="http://example.com" xmlns:b="http://example.com" a:bar="test" b:bar="toast"/>
  ```
  Maud currently happily accepts to output this, while it rightly refuses to produce something like
  ```
  <foo a="test" a="toast"/>
  ```
  But it's already usable. The user still needs to take some responsibility not to output invalid XML. This is also such an edge case that it should rarely impose a problem in practice. It might be more important to make sure that all used namespace prefixes are bound.
* Arbitrary doctypes besides html
* Processing instructions
* Possibly CDATA sections, but that's not really necessary because a consuming applications should not make a difference between text stored in CDATA and text that is ampersand escaped.

I assume that escaping is already working as required. Doctypes with local [element type declarations](https://www.w3.org/TR/xml/#elemdecls) or [entity declarations](https://www.w3.org/TR/xml/#sec-entity-decl) can probably be ignored.

th-we:
Thinking about namespace validation:  Here are examples demonstrating challenges:

```rust
use maud::html as xml;
use maud::html as subtree;
use maud::PreEscaped;

fn subtree() -> PreEscaped<String> {
    subtree! {
        subtree ns1:bar="test" ns2:baz="toast" {}
    }
}

fn main() {
    let markup1 = xml! {
    foo xmlns:ns1="NS:1"  {
            (subtree())
        }
    };
    println!("{}", markup1.into_string());


    let reusable_subtree = subtree();
    let markup2 = xml! {
[...]

PR #341, opened 2022-05-07
emit short tag on empty blocks

HookedBehemoth:
This changes the emited markup with tags like this: `use href="#page" {}` from `<use href="#page"></use>` to `<use href="#page"/>`.
Other tags this is useful for are e.g. video and all svg subtags.

### DIFF
--- a/maud/tests/basic_syntax.rs
+++ b/maud/tests/basic_syntax.rs
@@ -169,7 +169,7 @@ fn colons_in_names() {
 #[test]
 fn hyphens_in_element_names() {
     let result = html! { custom-element {} };
-    assert_eq!(result.into_string(), "<custom-element></custom-element>");
+    assert_eq!(result.into_string(), "<custom-element/>");
 }
 
 #[test]
@@ -177,7 +177,7 @@ fn hyphens_in_attribute_names() {
     let result = html! { this sentence-is="false" of-course {} };
     assert_eq!(
         result.into_string(),
-        r#"<this sentence-is="false" of-course></this>"#
+        r#"<this sentence-is="false" of-course/>"#
     );
 }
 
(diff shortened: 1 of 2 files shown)


zr40:
Unfortunately `<foo/>` is not correct [according to the spec](https://www.w3.org/TR/2012/WD-html-markup-20120329/syntax.html#syntax-elements):

> Start tags consist of the following parts, in exactly the following order:
> 1. A "<" character.
> 2. The element’s [tag name](https://www.w3.org/TR/2012/WD-html-markup-20120329/syntax.html#tag-name).
> 3. Optionally, one or more [attributes](https://www.w3.org/TR/2012/WD-html-markup-20120329/syntax.html#attribute), each of which [must](https://www.w3.org/TR/2012/WD-html-markup-20120329/terminology.html#must-requirement) be preceded by one or more [space characters](https://www.w3.org/TR/2012/WD-html-markup-20120329/terminology.html#space).
> 4. Optionally, one or more [space characters](https://www.w3.org/TR/2012/WD-html-markup-20120329/terminology.html#space).
> 5. Optionally, a "/" character, **which may be present only if the element is a [void element](https://www.w3.org/TR/2012/WD-html-markup-20120329/syntax.html#void-element).**
> 6. A ">" character.

The optional `/` character has no significance to HTML parsers, so it does not have the effect you desire. For instance, `<div/><div/>` is equivalent to `<div><div>`.

For void elements, those that do not require a closing tag, you can use `;` instead of `{}`.

HookedBehemoth:
Thanks for letting me know. Came from xmpp and just assumed this would be XML standard.

lambda-fairy:
I have a pending feature request for XML support at #62. It's a bit more complicated than it looks because of 🅰️ context-aware escaping and 🅱️ namespaces.

PR #345, opened 2022-05-30
Add ability to read html from file

alexkazik:
This allows for using external html, when it's a big chunk it's better to read when not part of the rust file.
```
html_file!("templates/base.maud");
```
It will be searched relative to the project root.

I've tested this only on top of https://github.com/lambda-fairy/maud/pull/322 and rocket 0.5 support, not on actual main but it should be of no difference.

### DIFF
--- a/maud/src/lib.rs
+++ b/maud/src/lib.rs
@@ -14,7 +14,7 @@ extern crate alloc;
 use alloc::{borrow::Cow, boxed::Box, string::String};
 use core::fmt::{self, Arguments, Write};
 
-pub use maud_macros::{html, html_debug};
+pub use maud_macros::{html, html_debug, html_file};
 
 mod escape;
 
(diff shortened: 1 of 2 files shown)


lambda-fairy (commented):
Hi @alexhazik, thanks for the PR!

This looks like a reimplementation of [`std::include!`](https://doc.rust-lang.org/std/macro.include.html).

In general, Maud aims to reuse built-in Rust features like modules and functions to organize its templates. So I prefer not to add features in that area.

For these reasons I will close this PR. Let me know if you have any follow up questions!

alexkazik:
`include` reads a file and places it in the code so it will be compiled as rust code.

`html_file` reads a maud template and returns a Html with the template.

And you could use include to do the same but would have to include html! in it, this looks as if it's made for it.

I've done this because it's similar to Haskell's Shakespeare, and they also have macros for in-line and in file.

alexkazik:
This PR has a little bug anyway.

But it would be a good idea to add an include! example to the website.

PR #358, opened 2022-10-28
Bump itoa to v1

paolobarbolini:
itoa 0.4 used `mem::uninitialized`, which is a very bad API as it's way of doing uninitialized memory goes against Rust's memory model. Starting from 1.64 it's slower because of the [forced memory initialization](https://github.com/rust-lang/rust/pull/99182/) it does to avoid the unsoundness


lambda-fairy:
Thanks!

Issue #266, opened 2021-03-19
impl Render for borrows ? 


RCasatta:
I struggled a bit understanding why this code doesn't work

```
#[test]
fn yet_another_test() {
    let mut vec = vec![];
    let a = html! { div { "foo" } };
    vec.push(a);
    let b = html! { div { "bar" } };
    vec.push(b);
    dbg!(&vec);

    let f = html! { @for el in &vec { (el) } };
    assert_eq!("<div>foo</div><div>bar</div>", f.into_string());
}
```

before realizing I had to Dereference el -> `*el`

Maybe it's possible to impl Render for borrowed values ?

lambda-fairy:
That's a good point.

I think it would be possible to add an impl for `&PreEscaped<T>`.

But I don't think we can generalize beyond that. An impl for all `&T` will conflict with the existing [blanket impl for `T: Display`](https://github.com/lambda-fairy/maud/blob/8990049101246aec0cf3c784182ca819c2c54ebc/maud/src/lib.rs#L72), because `Display` is implemented for all `&T` already.

I think the ideal solution is to remove the blanket impl for `T: Display`. This is what markup.rs and sailfish do. But we'll need to check how much code that breaks.

RCasatta:
> I think it would be possible to add an impl for &PreEscaped<T>.

this would be nice since it will cover most common case I think

lambda-fairy:
This is fixed by #320

PR #359, opened 2022-11-19
Render with Display using autoref specialization

simonask:
This may not be something you're interested in merging. No hard feelings if you close this outright. :-)

I found for my use case that I needed to write `display(...)` in a lot of places, especially in cases where I was using external types (especially the many datetime formatting types from `icu`).

This PR adds [autoref-based specialization](http://lukaskalbertodt.github.io/2019/12/05/generalized-autoref-based-specialization.html) to the `html!` macro such that it will choose a type's `Render` implementation if it is available, and use `Display` as a fallback.

The blanket impl of `Render for T where T: Display` was removed in #271 for good reasons, and I will try to justify this PR in terms of those reasons:

- #266 (Impl Render for borrows): This PR does not add any new blanket impls of `Render` or `Display`.
- Performance: Due to specialization, integers can still be rendered using `itoa`.
- Semantics: Arguably, `Display` is intended to provide a human-readable string. While this isn't necessarily tied to the HTML representation of the thing, it is not unreasonable to want to use the output of `Display` in HTML. This solution treats the output of `Display` as unescaped content (so it will be escaped before appending, by going through `maud::display(...)`). If a situation arises where the `Display` and `Render` implementations should be completely different, specialization allows this on a per-type basis.

The main drawback of this PR is that compiler errors are slightly more confusing when a type implements neither `Render` nor `Display`. I have tried to name the magic inner types used in the macro with that in mind.

Let me know what you think. :-)

### DIFF
--- a/maud/src/lib.rs
+++ b/maud/src/lib.rs
@@ -340,3 +340,54 @@ mod axum_support {
         }
     }
 }
+
+#[doc(hidden)]
+pub mod macro_private {
+    use crate::{display, Render};
+    use alloc::string::String;
+    use core::fmt::Display;
+
+    #[doc(hidden)]
+    #[macro_export]
+    macro_rules! render_to {
+        ($x:expr, $buffer:expr) => {{
+            use $crate::macro_private::*;
+            match ChooseRenderOrDisplay($x) {
+                x => (&&x).implements_render_or_display().render_to(x.0, $buffer),
+            }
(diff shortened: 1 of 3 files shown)


lambda-fairy:
Thanks and sorry for the delay!

I didn't expect the solution to be so clean – seeing your take validates the approach for me.

PR #337, opened 2022-04-03
Update `axum-core` depedency to 0.2

CBenoit:
Hi! Just updating optional dependency `axum-core`
(Thanks for the great crate!)

PR #361, opened 2022-11-25
Bump axum-core to 0.3.0

davidpdrsn:
This is required for axum 0.6 support

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 - Remove `html_debug!`
   [#357](https://github.com/lambda-fairy/maud/pull/357)
+- Bump axum-core to 0.3.0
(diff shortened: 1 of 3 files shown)


lambda-fairy (changes requested):
Thanks for the PR!

Would you like to update the changelog and doctests as well? You can use #337 as an example.

dsaghliani:
The doctest is failing because the `axum` dependency in `doctest/Cargo.toml` has to be updated to "0.6". Tweak that and it should work.

kotx:
Any estimate on when this may be ready?

davidpdrsn:
@lambda-fairy sorry for the delay. I've updated the changelog and fixed the doc tests now.

lambda-fairy (approved):

lambda-fairy:
Thanks!

PR #370, opened 2023-03-19
Add axum to the list of frameworks in docs

drdo:
Axum was missing in the list at the top of the file.

Not sure if the other frameworks should be added to `index.md`,
only Rocket and Actix are mentioned there.


lambda-fairy:
Thanks!

PR #371, opened 2023-03-24
impl Default for PreEscaped

imbolc:
Tests was broken before the change, I'm not sure if I should fix them. But `cargo test misc` works.

### DIFF
--- a/maud/tests/misc.rs
+++ b/maud/tests/misc.rs
@@ -130,3 +130,10 @@ fn prefer_render_over_display() {
         "&lt;display&gt;"
     );
 }
+
+#[test]
+fn default() {
+    use maud::{Markup, PreEscaped};
+    assert_eq!(Markup::default().0, String::from(""));
+    assert_eq!(PreEscaped::<&'static str>::default().0, "");
+}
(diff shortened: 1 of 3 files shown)


imbolc:
Also I noticed seemingly unnecessary type restriction: 
`struct PreEscaped<T: AsRef<str>>(pub T)`. I believe removing the `AsRef` restriction won't change any behavior, but it won't be necessary to repeat the restriction in each implementation. I can remove it in this PR if you'd like.

maud/tests/misc.rs
```diff
@@ -130,3 +130,10 @@ fn prefer_render_over_display() {
         "&lt;display&gt;"
     );
 }
+
+#[test]
+fn default() {
+    use maud::{Markup, PreEscaped};
+    assert_eq!(Markup::default().0, String::from(""));
```
lambda-fairy, line 137: Nit: `String` can be compared with `&str`. So this works:

```suggestion
    assert_eq!(Markup::default().0, "");
```

lambda-fairy:
Thanks!

> I believe removing the `AsRef` restriction won't change any behavior, but it won't be necessary to repeat the restriction in each implementation. I can remove it in this PR if you'd like.

That sounds like a good change. Though I'd prefer that in a separate PR, just to keep things small.

That reminds me, I should polish up #322 again...

PR #377, opened 2023-04-16
Remove `AsRef<str>` restriction from structs

imbolc:
Following #371

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+- Remove `AsRef<str>` restriction from `PreEscaped`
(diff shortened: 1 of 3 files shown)


lambda-fairy:
Thanks!

Would you like to add a [changelog entry](https://github.com/lambda-fairy/maud/blob/main/CHANGELOG.md) too? See the [contributor guide](https://github.com/lambda-fairy/maud/blob/main/CONTRIBUTING.md).

I added it myself last time, since I was cutting a release (and so updating the changelog anyway). But in general I'd like the author of the PR to do that.

lambda-fairy:
Thanks!

PR #393, opened 2023-10-10
Allow many ':' in attributes

RedPhoenixQ:
To do this, `try_namespace_name` now looks for nested namespaces. Because `try_name`'gets called, another `Ident `must be consumed to prevent and early return, which means that `::` would not be valid, but `a:b-c:d` would be fine.

This allows for more custom attributes from different frameworks, specifically htmx's `hx-on` as mentioned in  https://github.com/lambda-fairy/maud/issues/194#issuecomment-1734354727.

This is my first PR, I'll gladly take critique and make any changes necessary

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   [#377](https://github.com/lambda-fairy/maud/pull/377)
 - Implement `Render` for `Arc<T>`
   [#380](https://github.com/lambda-fairy/maud/pull/380)
+- Accept literals in attribute names
+  [#396](https://github.com/lambda-fairy/maud/pull/396)
 
 ## [0.25.0] - 2023-04-16
 
--- a/maud/tests/basic_syntax.rs
+++ b/maud/tests/basic_syntax.rs
@@ -182,6 +182,34 @@ fn hyphens_in_attribute_names() {
 }
 
 #[test]
+fn string_literals_in_attribute_names() {
+    let result = html! { this "@sentence:-is.not"="false" of-course {} };
+    assert_eq!(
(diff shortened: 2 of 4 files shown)


fraschm1998:
Could you add support for alpine.js as well for '.':

`x-on:scroll.window="showBar = (window.pageYOffset > 20) ? true : false"`

RedPhoenixQ:
> Could you add support for alpine.js as well for '.':
> 
> `x-on:scroll.window="showBar = (window.pageYOffset > 20) ? true : false"`

I don't think this is easily done while keeping the class shorthand. Since the macro doesn't see whitespace `foo.bar` is the same as `foo .bar` which would be a valid attribute and then a class name via the shorthand. 

The only way to do this would to somehow replace or escape the period `.` in the attribute name or add an escape character that would have to be a recognized Rust token. `\` cannot be used as it isn't part of Rust syntax. Something like `*.` or `^.` could work but isn't very elegant.

I don't want to make a change like class shorthand can only appear immediately after the element since that would be a breaking change.

fraschm1998:
Any idea how difficult it would be to add quoted attribute supported as mentioned [here](https://github.com/lambda-fairy/maud/issues/194#issuecomment-1527303886)

RedPhoenixQ:
Though about it and just tried it. I seem to have it working and will be making a new PR for that change, since it is not related to the changes in this PR.

RedPhoenixQ:
#396 Now fixes the same issue by allowing `"literals"="true"`

dacbd:
Would like to see this feature too 👍

Issue #194, opened 2020-06-15
Allow attributes with special characters


edg-l:
```
body {
    button @click="open = true" { "Open" }
}
```

I think you should be able to somehow use "@click" as an attribute name, this comes in handy when using a library like [alpinejs](https://github.com/alpinejs/alpine)

lambda-fairy:
Hi @edg-l!

According to the [Alpine.js docs](https://github.com/alpinejs/alpine/blob/master/README.md#x-on), you can use:

```
button x-on:click="open = true" { "Open" }
```

As mentioned in #134, I'd rather not add special syntax to support languages other than HTML. So I'll go ahead and close this. Let me know if you have any other questions!

charlesvdv:
Hello @lambda-fairy,

I just stumbled another _incompatibility_ between maud and alpine. In alpine, you can use [modifiers](https://alpinejs.dev/directives/on#modifiers) which customise the event handler. For example, you can do `x-on:click.outside="foo()"` to call `foo()` when the user click outside of the DOM element. 

In Maud, the dot character is used to denote a class. Using modifiers in maud templates leads to syntax errors.

I'm thinking maud could support quoted attribute name like this: 

```
div "x-on:click.outside"="foo()" "@click"="bar()" {}
``` 

As you can see this could as well solve the problem explained by @edg-l.

What do you think? If looks ok to you, I could look into integrating the proposal in maud.

Thanks a lot!

spiffytech:
As another data point, HTMX [is working on](https://htmx.org/extensions/response-targets/) a very useful extension that uses numbers and asterisks in element attributes.

Its usage looks like this:

```html
<button hx-post="/register"
            hx-target="#response-div"
            hx-target-5*="#serious-errors"
            hx-target-404="#not-found">
        Register!
    </button>
```

It looks like the `*` isn't valid HTML, but numbers appear to be valid characters for attributes, which Maud doesn't accept right now.

I'm unsure how to proceed in cases like this. Should I fall back to stuffing a `PreEscaped` HTML string in the middle of my template?

Would you consider reopening this? @charlesvdv's proposed syntax appears to solve this.

josdelien:
> Hello @lambda-fairy,
> 
[...]

PR #396, opened 2023-10-19
Accept literals in attribute names

RedPhoenixQ:
The parser now accepts `TokenTree::Literal` as names for attributes.

To prevent the `name_to_string` function from showing and escaping the surrounding `"` of a string literal the string representation is trimmed first. This should not affect how other literals work which means that both `"name"="true"` and `123="true"` would be valid.

This also addresses the PR request in #194

### DIFF
--- a/maud/tests/basic_syntax.rs
+++ b/maud/tests/basic_syntax.rs
@@ -182,6 +182,15 @@ fn hyphens_in_attribute_names() {
 }
 
 #[test]
+fn string_literals_in_attribute_names() {
+    let result = html! { this "@sentence:-is.not"="false" of-course {} };
+    assert_eq!(
+        result.into_string(),
+        r#"<this @sentence:-is.not="false" of-course></this>"#
+    );
+}
+
+#[test]
 fn class_shorthand() {
     let result = html! { p { "Hi, " span.name { "Lyra" } "!" } };
     assert_eq!(
--- a/maud/tests/basic_syntax.rs
+++ b/maud/tests/basic_syntax.rs
@@ -191,6 +191,25 @@ fn string_literals_in_attribute_names() {
 }
 
 #[test]
+fn raw_string_literals_in_attribute_names() {
+    let result = html! { this r#"@sentence:-is.not"#="false" of-course {} };
+    assert_eq!(
+        result.into_string(),
+        r#"<this @sentence:-is.not="false" of-course></this>"#
+    );
+}
+
+#[test]
+fn other_literals_in_attribute_names() {
+    let result =
+        html! { this b"byte_string"="false" 123="123" 2.5 true 'a'="a" b'b'="b" of-course {} };
+    assert_eq!(
+        result.into_string(),
+        r#"<this byte_string="false" 123="123" 2.5 true a="a" b="b" of-course></this>"#
+    );
+}
+
+#[test]
 fn class_shorthand() {
     let result = html! { p { "Hi, " span.name { "Lyra" } "!" } };
     assert_eq!(
--- a/maud_macros/src/ast.rs
+++ b/maud_macros/src/ast.rs
@@ -217,5 +217,13 @@ pub fn join_ranges<I: IntoIterator<Item = SpanRange>>(ranges: I) -> SpanRange {
 }
 
 pub fn name_to_string(name: TokenStream) -> String {
-    name.into_iter().map(|token| token.to_string()).collect()
+    name.into_iter()
+        .map(|token| {
+            if let TokenTree::Literal(literal) = token {
+                literal.to_string().trim_matches('"').to_string()
+            } else {
+                token.to_string()
+            }
+        })
+        .collect()
 }
--- a/maud_macros/src/ast.rs
+++ b/maud_macros/src/ast.rs
@@ -1,5 +1,6 @@
 use proc_macro2::{TokenStream, TokenTree};
 use proc_macro_error::SpanRange;
+use syn::Lit;
 
 #[derive(Debug)]
 pub enum Markup {
@@ -220,7 +221,15 @@ pub fn name_to_string(name: TokenStream) -> String {
     name.into_iter()
         .map(|token| {
             if let TokenTree::Literal(literal) = token {
-                literal.to_string().trim_matches('"').to_string()
+                match Lit::new(literal.clone()) {
+                    Lit::Str(str) => str.value(),
+                    Lit::Char(char) => char.value().to_string(),
+                    Lit::ByteStr(byte) => {
+                        String::from_utf8(byte.value()).expect("Invalid utf8 byte")
+                    }
+                    Lit::Byte(byte) => (byte.value() as char).to_string(),
+                    _ => literal.to_string(),
+                }
             } else {
                 token.to_string()
             }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   [#377](https://github.com/lambda-fairy/maud/pull/377)
 - Implement `Render` for `Arc<T>`
   [#380](https://github.com/lambda-fairy/maud/pull/380)
+- Accept literals in attribute names
+  [#396](https://github.com/lambda-fairy/maud/pull/396)
 
 ## [0.25.0] - 2023-04-16
 
--- a/maud_macros/src/parse.rs
+++ b/maud_macros/src/parse.rs
@@ -702,12 +702,13 @@ impl Parser {
     /// Parses an identifier, without dealing with namespaces.
     fn try_name(&mut self) -> Option<TokenStream> {
         let mut result = Vec::new();
-        if let Some(token @ TokenTree::Ident(_)) = self.peek() {
-            self.advance();
-            result.push(token);
-        } else {
-            return None;
-        }
+        match self.peek() {
+            Some(token @ TokenTree::Ident(_)) | Some(token @ TokenTree::Literal(_)) => {
+                self.advance();
+                result.push(token);
+            }
+            _ => return None,
+        };
         let mut expect_ident = false;
         loop {
             expect_ident = match self.peek() {


fraschm1998:
Tried this, I seem to get:

`"x-on:click"="toggle" type="button" aria-controls="mobile-menu" aria-expanded      ="false" {     ■ element body must be wrapped in braces      = help: see https://github.com/lambda-fairy/maud/pull/137 for details`

RedPhoenixQ:
Could you send more context like the macro where this element is created. This looks like and issue with missing `{}` for the element body or `;` for a self closing element

fraschm1998:
> Could you send more context like the macro where this element is created. This looks like and issue with missing `{}` for the element body or `;` for a self closing element

```
div."absolute inset-y-0 left-0 flex items-center sm:hidden" {                                                                                                                                                                                                                        
    1                                             // Mobile menu button                                                                                                                                                                                                                                                            
💥  2                                             button."relative inline-flex items-center justify-center rounded-md p-2 text-gray-400 hover:bg-gray-100 hover:text-gray-500 focus:outline-none focus:ring-2 focus:ring-inset focus:ring-indigo-500" "@click"="toggle" type="button" aria-controls="mobile-menu" aria-expanded="fa      lse" {     ■ element body must be wrapped in braces      = help: see https://github.com/lambda-fairy/maud/pull/137 for details                    
    3                                                 span."absolute -inset-0.5" {}                                                                                                                                                                                                                  
    4                                                 span."sr-only" { "Open main menu" }                                                                                                                                                                                                            
    5                                                 // Icon when menu is closed.                                                                                                                                                                                                                   
    6                                                 @for (item, path) in [("block", "M3.75 6.75h16.5M3.75 12h16.5m-16.5 5.25h16.5"), ("hidden", "M6 18L18 6M6 6l12 12")] {                                                                                                                         
    7                                                     svg.{ (item) " h-6 w-6" } fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" aria-hidden="true" {                                                                                                                                                  
    8                                                         path stroke-linecap="round" stroke-linejoin="round" d=(path) {}                                                                                                                                                                                                      
    9                                                     }                                                                                                                                                                                                                                                                        
   10                                                 }                                                                                                                                                                                                                                                                            
   11                                             }                                                                                                                                                                                                                                                                                
   12                                         } 
   ```

**Comment by RedPhoenixQ** [comment]
I'm unable to reproduce the issue when copying the elements you sent. On the code that includes this PR it compiles for me and seemingly give the correct html.

This is the html string your code produces:
`"<div class=\"absolute inset-y-0 left-0 flex items-center sm:hidden\"><button class=\"relative inline-flex items-center justify-center rounded-md p-2 text-gray-400 hover:bg-gray-100 hover:text-gray-500 focus:outline-none focus:ring-2 focus:ring-inset focus:ring-indigo-500\" @click=\"toggle\" type=\"button\" aria-controls=\"mobile-menu\" aria-expanded=\"false\"><span class=\"absolute -inset-0.5\"></span><span class=\"sr-only\">Open main menu</span><svg class=\"block h-6 w-6\" fill=\"none\" viewBox=\"0 0 24 24\" stroke-width=\"1.5\" stroke=\"currentColor\" aria-hidden=\"true\"><path stroke-linecap=\"round\" stroke-linejoin=\"round\" d=\"M3.75 6.75h16.5M3.75 12h16.5m-16.5 5.25h16.5\"></path></svg><svg class=\"hidden h-6 w-6\" fill=\"none\" viewBox=\"0 0 24 24\" stroke-width=\"1.5\" stroke=\"currentColor\" aria-hidden=\"true\"><path stroke-linecap=\"round\" stroke-linejoin=\"round\" d=\"M6 18L18 6M6 6l12 12\"></path></svg></button></div><h1 class=\"pinkie-123 asdf asdf asdf asdf asdf\">Pinkie Pie<button class=\"asdf\">test</button></h1>"`

**Comment by fraschm1998** [comment]
Nvm. Seems to be working now. I think I may have just forgotten to clear the build cache.

Thanks!

**Comment by RedPhoenixQ** [comment]
Are you using this PR locally? In your `Cargo.toml` you can replace `maud = X.X.X` with `maud = { git = "https://github.com/RedPhoenixQ/maud/tree/literals_in_attributes" }` and that should let you compile your code

File: maud_macros/src/ast.rs [hunk 1]
```diff
@@ -217,5 +217,13 @@ pub fn join_ranges<I: IntoIterator<Item = SpanRange>>(ranges: I) -> SpanRange {
 }
 
 pub fn name_to_string(name: TokenStream) -> String {
-    name.into_iter().map(|token| token.to_string()).collect()
+    name.into_iter()
+        .map(|token| {
+            if let TokenTree::Literal(literal) = token {
+                literal.to_string().trim_matches('"').to_string()
```
lambda-fairy, line 223: Does `.trim_matches("")` work if the literal is a raw string, or contains escapes?

lambda-fairy (changes requested):
Thanks @RedPhoenixQ, looks great so far!

Can you please add a changelog entry as well?

maud_macros/src/ast.rs
failingprovince: Yes.
Check out [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cecce2b4d6f655546d41fe8650d42add) playground.

RedPhoenixQ: Is this the wanted behaviour? That the raw string `r#""test""#` should be output as just `test` and not `"test"` as the raw string would imply?

Either way it can easily be changed

RedPhoenixQ: Strings like `r#""t"es"t""#` would still include the interior `"` like `t"es"t`, se  [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=36a4d54383c8eb412ce7db244b784da8) modified playground

failingprovince: Well `trim_matches()" does exactly this: see the [doc](https://doc.rust-lang.org/stable/std/primitive.str.html#method.trim_matches).
If this isn't the wanted behaviour then why use this function in first place.

Still I see your point, I just made the assumption that if `trim_matches()` was used then this behaviour was actually correct.

RedPhoenixQ: This i the behavior I imagined since attributes starting with `"` does not seem relevant, but I do not intend to force my view of the behavior on the project. If attributes starting or ending in `"` was needed I would change the implementation

failingprovince: My bad. Haven't read the previous implementation and thought that `trim_matches` was already used beforehand.

lambda-fairy: My concern was that `r#"test"` would be output as literally `r#"test"`. Because `proc_macro::Literal::to_string()` returns the **source code**, not the final string.

If you add the following test then it should be clear what the problem is.

```rust
#[test]
fn string_literals_in_attribute_names() {
    let result = html! { this r#"@sentence:-is.not"#="false" of-course {} };
    assert_eq!(
        result.into_string(),
        r#"<this @sentence:-is.not="false" of-course></this>"#
    );
}
```

Take a look at the [literal parser](https://github.com/lambda-fairy/maud/blob/d1cfde53b6845c0b58c49b77167a5ce27cc7f11b/maud_macros/src/parse.rs#L198) for ideas on how to handle this correctly.

RedPhoenixQ: > proc_macro::Literal::to_string() returns the source code, not the final string.
Thank you for the clarification. Fixed in c33ee62d87f7d0d2840c0b124ac424f1c311d353

maud/tests/basic_syntax.rs
```diff
@@ -181,6 +181,34 @@ fn hyphens_in_attribute_names() {
...
+    );
+}
+
+#[test]
+fn other_literals_in_attribute_names() {
+    let result =
+        html! { this b"byte_string"="false" 123="123" 2.5 true 'a'="a" b'b'="b" of-course {} };
+    assert_eq!(
+        result.into_string(),
+        r#"<this byte_string="false" 123="123" 2.5 true a="a" b="b" of-course></this>"#
```
lambda-fairy, line 208: I think string literals, and _maybe_ integers, are enough.

Floating point numbers overlap with class shorthand which also uses `.`. There's no use case for a single character. And no other part of Maud accepts byte strings.

lambda-fairy (approved):
Thanks @RedPhoenixQ!

Let's merge what we have now. We can address my other comment in a follow-up PR.

maud/tests/basic_syntax.rs
RedPhoenixQ: @lambda-fairy, addressing this issue, to keep `ast::name_to_string()` as an infallible api (always returning `String` without errors) I handled all the cases since I didn't see any issue with also handling `ByteStr` and others so they produce the expected internal string. Current implementation should always produce valid html.

The class shorthand shouldn't be broken since classes starting with numbers are disallowed in CSS so all classes before should still be possible. Also the only time a class starting with a number wouldn't be interpreted as a class would be ` 123.123` (space followed by numbers with dot and no spaces), but `123 .123`, `anything.123` and `.classname123.123` would still parse as a class shorthand for  the class `123`.

RedPhoenixQ:
I found a bug. Literals are now parsed and the `try_name()` function doesn't see literals, leading to code like `html! { div test-123="true" {} }` to be generated as `<div test- 123="true"></div>`

maud/tests/basic_syntax.rs
lambda-fairy: Having the interpretation change just by adding a space is the confusing part.

RedPhoenixQ: #398 fixes this issue by running the existing `literal()` function in the parser when encountering a literal, reusing existing error behaviour. I've added `Int` as an accepted literal, but can revoke that change if you decide that `Int` shouldn't be allowed.

PR #336, opened 2022-03-29
Support for classes containing numbers in their names

JohnDowson:
Closes #335

Issue #335, opened 2022-03-28
It is impossible yo use classes containing numbers.


JohnDowson:
Notably Bootstrap has alot of classes like `mx-4`, `px-3`, `gy-2`.

lambda-fairy:
Hi @JohnDowson, thanks for the suggestion and PR!

My primary concern with this (and the reason why numbers weren't permitted originally) is that in Rust proper, `0000069` and `69` are considered to be the same number. But if we allow your proposal, then Maud will introduce a distinction between the two. I'm unsure what implications this might have.

In particular:

1. Does Rust itself guarantee that number literal tokens are passed through unchanged to proc macros? (So that the token `0000069` is _not_ normalized to `69` or anything else.)
2. What about other tools, like IDEs or code formatters?
3. Finally, from a user's point of view, would they be confused by this?

Also, consider that there is already an alternative syntax (`."mx-4"`). See #267.

puetzp:
For me this is more of a documentation issue than a matter of usability. However since I hit this problem myself and found this issue to resolve it, the [documentation](https://maud.lambda.xyz/elements-attributes.html) was updated and explains how to the "alternative syntax" in order to use numbers in class names.
From my point of view there is no need to spend (probably) a lot of time to evaluate the implications mentioned above.

trepidacious:
I ran into a similar issue with tailwind, so far my preferred approach is to just use `class="some-class"` or `class=(splice)`. I don't think this is mentioned in the docs but seems to work fine? One thing I did notice is that if I use `.class1 .{(splice)}` I will get both `class1` and the spliced classes in the resulting html, using `class="class1" class=(splice)` seems to miss out `class1`?

This way I don't need to think about quoting only the class names that need it. Another advantage is that the vs code tailwind plugin just needs rust files to be added to the list of relevant languages and it will then pick up the class names. If I use the `.class-name` syntax the plugin needs an experiment regex setting to find class names, and it's pretty tricky to find a regex that covers all the options for Maud classes without also picking up false positives on method calls in normal rust code. Finally it makes the class names a little easier to pick out since they're highlighted as strings.

It's a shame because I do think it's cool to be able to have `.class-name`, especially where there are only one or two classes.

PR #398, opened 2023-11-14
Parse many literals along side idents in names

RedPhoenixQ:
This is a fix for the bug mentioned in https://github.com/lambda-fairy/maud/pull/396#issuecomment-1811008846. 

This is similar to #336 and also addresses #335, se the test for examples of what is now possible

The implications mentioned in https://github.com/lambda-fairy/maud/issues/335#issuecomment-1094247196 are not addressed here, but for now this fixes the bug that #396 created


lambda-fairy (approved):
Thanks for the fix!

PR #400, opened 2023-11-25
Upgrade to `syn` 2.0.0

sgued:
This PR bumps the `syn` dependency to version 2.0.0

It also removes the default features of `proc-macro-error` (which weren't used) to remove the transitive dependency on `syn` 1.0.0.


lambda-fairy:
Thank you!

PR #404, opened 2023-12-02
Warp Integration: Yet another web framework support

thatguydoru:
This implements `warp::Reply` for `Markup`, making it possible to return it from a handler and making *maud* more pleasing to use with *warp*.

### DIFF
--- a/maud/Cargo.toml
+++ b/maud/Cargo.toml
@@ -27,6 +27,7 @@ actix-web-dep = { package = "actix-web", version = "4", optional = true, default
 tide = { version = "0.16.0", optional = true, default-features = false }
 axum-core = { version = "0.3", optional = true }
 http = { version = "0.2", optional = true }
+warp = { version = "0.3.6", optional = true }
 
 [dev-dependencies]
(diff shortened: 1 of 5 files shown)


maud/Cargo.toml
```diff
@@ -27,6 +27,7 @@ actix-web-dep = { package = "actix-web", version = "4", optional = true, default
 tide = { version = "0.16.0", optional = true, default-features = false }
 axum-core = { version = "0.4", optional = true }
 http = { version = "1", optional = true }
```
lambda-fairy, line 29: Hmm, looks like Axum uses http 1.0, but Warp uses http 0.2. But since the Warp integration doesn't use the http dependency directly, it still works out? Still something to look out for in the future.

lambda-fairy:
Thanks!

The `http` dependency issue looks ok for now, so let's merge this and revisit if it causes problems later.

maud/Cargo.toml
thatguydoru: Noted, I'll have a close eye on it. 🙇

Issue #386, opened 2023-07-27
Support adding attributes through iterators


jquesada2016:
Please foregive me if this is already possible, as I did not find it in the book. But I am trying to componentize some markup, and need to support adding arbitrary attributes. So for example, let's take an `<input>`

```rust
struct Input {
  attrs: Vec<(String, String)>
}

impl Render for Input {
  fn render(&self) -> Markup {
    html! {
      input type="text" ...self.attrs;
    }
  }
}
```

Is something like this currently possible? I understand the security implications, but this is intended to be done at compile time, by adding attributes in different parts of the app, not dynamically.

jquesada2016:
If not currently possible, perhaps iterators and a spread operator might be good additions to the syntax?

Attribute spreading needs to implement `Iterator<Item = (impl Display, impl Display)> and could use the rust spread operator `..` while classes can be spread via `. ..` or something similar? Would only need to implement `Iterator<Item = impl Display>`

Issue #240, opened 2020-11-26
dynamic attributes ideas


Brecert:
Currently only the value of an attribute can have a dynamic value, having the ability to not just set the value but also the attribute name/type would be useful.

What I mean when talking about dynamic attributes is shown in the below examples.
As far as I know, other than the first, none of these are really possible at the moment.

- a way to have the attribute value able to be modified (exists)
  ```rust
  let value = "value";
  html! { span title=(value) }
  // <span title="value">
  ```
- a way to have the attribute name able to be modified
  ```rust
  let attribute_name = "title";
  html! { span (attribute_name)="value"
  // <span title="value">
  ```
- a way to enable/disable adding said attribute with types that implement `Try`
  ```rust
  html! { span (Some(attribute_name))?="value" } 
  // <span title="value">

  html! { span (None)?="value"  }
  // <span>
  ```
- a way to destructure attributes, through an Iterator that iterates with a tuple that matches `(K, V)`
  ```rust
  // maybe with hashmap or someth
  let attributes = HashMap::new();
  attributes.insert("title", "value");
  attributes.insert("alt", "a value");
  html! { img (..)=(attributes)  }
  // <img title="value" alt="a value">
  ```

lambda-fairy:
Thanks for the feedback!

I've been holding off on making the attribute name dynamic, because it complicates context-aware escaping (#181).

Do you have a concrete use case for these features? That'll help us shape what they'll look like.

Brecert:
I'm currently writing a markup language where certain properties are represented by html attributes.

Conditionally adding an attribute is useful for me as certain attributes behave differently between not existing, being a boolean, and having an empty string.

It's also beneficial to me to produce clean templates and html.

Destructuring is useful for me because the language and generation is meant to be extendable, so having a way to add attributes for the extended parts dynamically is important.

Zopieux:
@brecert One concrete (and I believe very common?) use-case that today I don't see how to express in Maud is the dynamic *presence* of attributes, which is suggestion 3 from OP, ie. either A. `span attr=(value) { }` or B. `span { }`, where say `value` is `Option<String>`. Currently I need to either always have `attr` with an empty/non-empty value, which as you noted is different from **not setting** `attr`, which is what I want. Sure, I could have a verbose `@if` or `@match` and duplicate the entire element declaration, but that looks sub-optimal.

[...]

PR #408, opened 2023-12-18
Support attribute spreading with iterators

JonahLund:
This PR adds the ability to spread attributes using iterators (#386, #240).

The syntax is as following:

`(..)=[attrs]` Where `attrs` is an expression that supports any iterator with tuples where the name and value implements `Display`.


jquesada2016:
Did something happen with this PR?

PR #409, opened 2023-12-26
Axum API change

OrganicProgramming:
Axum-core and HTTP API have slightly changed for one of the latest Axum releases and the intoResponse trait was no longer satisfied. This fix works again for the latest release of Axum, but the author of this library might choose to support the older API as well. The public API of Maud is not affected by this fix.


OrganicProgramming:
Fix already available in one of the other pull requests.

PR #421, opened 2024-02-20
Add trait implementation for Vec<T>, [T], Option<T>

nik-k999:


nik-k999:
The CI on the main branch seems to fail generally. [#1556](https://github.com/lambda-fairy/maud/actions/runs/7967346681). So this Pull request will probably be blocked until the CI errors on main branch are fixed.

14ROVI:
This looks good, and super useful. Hopefully it can be merged soon :D

sorcix:
I actually like that `Option<T>` isn't supported out of the box. Usually when I encounter the error that an `Option<T>` can't be rendered, it is a mistake on my side, and more than just the value has to be hidden when it is `None`. (e.g. enclosing markup)

The implementation in this MR would just hide the value.

Issue #446, opened 2024-10-09
impl Render<T: Render> for Option<T>


martinhath:
Thanks for a nice library! 

I ran into a case where I want to conditionally insert some markup, the condition being on an `Option`. Ended up writing 
```rs
html! {
    @if let Some(n) = foo() { (n) }
}
```
but it would have been cool if 
```rs
html! {
    (foo())
}
```
worked, and behaved identically.

re-oh:
Ill attempt to fix this, want to contribute since this is a really cool lib.

re-oh:
```rust
impl<T: Render + ?Sized> Render for Option<T> {
    fn render_to(&self, w: &mut String) { if let Some(inner) = self { T::render_to(inner, w); } }
}

impl <T: Render + ?Sized, I: ExactSizeIterator + ?Sized> Render for I where I::Item: AsRef<T> {
    fn render_to(&self, w: &mut String) {
        for item in self {
            item.as_ref().render_to(w);
        }
    }
}
```

Implemented Render for 2 types:
- `ExactSizeIterator` since if it was infinite like: `0..` it would run forever. ( Not sure if this implementation works, if anyone could tell me how i can test this it would be much appreciated since you dont need a owned value to render it. ergo `AsRef<T>` for `I::Item`)
- `Option<T>` Simple, if self is some render the inner value if not, dont render it. ( I think thats how @martinhath wanted it to work. )

If there are any suggestions for other types to implement or if my implementation has issues ill gladly fix them.

Porges:
This would be nice, it’s similar to how both `null`/`false` skip rendering in React.

PR #447, opened 2024-10-10
Added: Render for Option<T>

re-oh:
```rust
impl<T: Render + ?Sized> Render for Option<T> {
    fn render_to(&self, w: &mut String) { if let Some(inner) = self { T::render_to(inner, w); } }
}

impl <T: Render + ?Sized, I: ExactSizeIterator + ?Sized> Render for I where I::Item: AsRef<T> {
    fn render_to(&self, w: &mut String) {
        for item in self {
            item.as_ref().render_to(w);
        }
    }
}
```

Implemented Render for 2 types:
- `ExactSizeIterator` since if it was infinite like: `0..` it would run forever. ( Not 100% sure if this implementation works, if anyone could tell me how i can test this it would be much appreciated since you dont need a owned value to render it. ergo `AsRef<T>` for `I::Item`)
- `Option<T>` Simple, if self is some render the inner value if not, dont render it. ( I think thats how @martinhath wanted it to work. )

If there are any suggestions for other types to implement or if my implementation has issues ill gladly fix them.

### DIFF
--- a/maud/src/lib.rs
+++ b/maud/src/lib.rs
@@ -156,14 +156,10 @@ impl<T: Render + ?Sized> Render for Arc<T> {
     }
 }
 
-impl<T: Render + ?Sized> Render for Option<T> {
-    fn render_to(&self, w: &mut String) { if let Some(inner) = self { T::render_to(inner, w); } }
-}
-
-impl <T: Render + ?Sized, I: ExactSizeIterator + ?Sized> Render for I where I::Item: AsRef<T> {
+impl<T: Render> Render for Option<T> {
     fn render_to(&self, w: &mut String) {
-        for item in self {
-            item.as_ref().render_to(w);
+        if let Some(inner) = self {
+            T::render_to(inner, w);
         }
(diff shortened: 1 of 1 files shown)


re-oh:
Also removed the `?Sized` bound on `T` for `Option<T>` since Apparently `Option<T>` needs for `T` to be sized.

re-oh:
closes issue / feature request: #446

eldad:
I had the same wish and now I see this is already been requested.
This would be a nice QoL improvement. Makes sense to include a simple test with this feature (https://github.com/eldad/maud/commit/e734169308ef6b0d40950cbd862e72b3bf44d870).

eldad:
On second thought, unbound nesting is probably not desired. 

Perhaps a marker trait can be used to restrict it to only the types Render is already implemented for.

For example: https://github.com/eldad/maud/commit/0d2dab9041f597c14661c6d395a400407bd03a55
(incomplete; does not cover some types)

PR #450, opened 2024-11-12
Abort when `else` is used without `@` (#449)

### LINKED ISSUE #449: prevent accidental misuse of if-else
i had this template in my code:

```
@if condition { path_a }
else { path_b }
```

maud rendered this to:

```
path_a
<else>path_b</else>
```

it would be nice if maud could make all of its own constructs reserved keywords, so that this situation doesn't occur anymore. there shouldn't be a reason to ever create an `<else>` tag. if there is, surely new syntax like `"else" { .. }` can satisfy that usecase.

puetzp:
Fixes #449.


lambda-fairy:
Thanks!

Issue #392, opened 2023-09-30
Feature: Hot-reload / Runtime interpretor for debug mode


wrapperup:
Love the library, it's really nice!

This would add an optional feature that allows maud to embed a runtime parser/interpreter that can read its own macro block from Rust source code. This runtime templating is limited in scope, meaning that changes to data, bindings, etc will require a recompile, but other changes could be done by interpreting the macro block at runtime (which IMO covers a lot of iteration time). This is definitely a bigger task with multiple steps, but I think it can be a really useful feature.

This is something Dioxus has implemented as a CLI tool, and I made a proof-of-concept implementation `html_hotreload!` (renamed to `html!` in the video) macro that uses maud's parser to generate a runtime instead that also parses itself at runtime (inception...).

https://github.com/lambda-fairy/maud/assets/7478134/274f2fcd-99e2-4cc2-899d-925d07f08263

Is this something you'd be interested in including with maud as an optional feature, or is it out of scope for maud? I think it could be it's own crate with a small refactor to expose maud's proc-macro/parser internals.

Either way, thanks for the useful library!

lambda-fairy:
Hi @wrapperup, thanks for sharing! That looks really cool.

I think I'm too busy right now to review major changes like this. Feel free to iterate out-of-tree!

untitaker:
i'd definitely be interested in a version of this that works with axum and my own htmx frontend, but i can't figure out which of the dioxus crates offers this functionality.

untitaker:
Found the code: https://github.com/wrapperup/maud/commits/hotreload-prototype/

wrapperup:
> Found the code: https://github.com/wrapperup/maud/commits/hotreload-prototype/

Yeah, the code there is nowhere near finished and it pulls in some extra unneeded dependencies, so I'm not sure if it would be a good idea to use it in the current state, but feel free to use it as a reference :).

If this work is something you're interested in continuing, the main blocker is splitting out maud's proc-macro internals into a different crate, making a bit of it public and upstreaming that (you can see some of the work splitting `maud_macros` out into `maud_macros_impl`). Ideally, a crate that implements the hot-reload functionality would just depend on `maud_macros_impl` and be a separate module/crate.

untitaker:
fixed merge conflicts with maud 0.26: https://github.com/untitaker/maud/pull/new/hotreload-prototype-v2

it works perfectly for me as-is. just swap `maud::html!` with `maud::html_hotreload!` as you said.

[...]

PR #451, opened 2024-11-17
Hotreload: Ability to update templates without recompiling

untitaker:
This is a continuation of #392, it adds a way to auto-reload templates to maud without going through rustc at all, for **some situations.**

## Basic usage

Enable the `hotreload` feature, and the `html!` macro will partially evaluate at runtime. use `maud::html_static!` to force the "normal", fully-compiled behavior.

You can use `export MAUD_ON_ERROR=exit` to make maud call `std::process::exit` during template-rendering whenever it turns out the code changes are too severe for hot-reloading. In combination with `cargo watch`, this can hot-reload your web app's templates when possible, and recompile when that fails:

```
MAUD_ON_ERROR=exit  cargo watch --no-restart -x run --features hotreload
```

This has some limitations because now rust code changes unrelated to templates will not cause the application to restart. Needs further improvements.

## Internals

maud+hotreload will compile your template into two parts:

1. a list of "formatting arguments" where all the rust code in templates is eagerly evaluated
2. a runtime that fetches the most recent template and attempts to use the existing "formatting arguments" with that

This means that changes such as this can be hot-reloaded:

```rust
// before
html! {
    "Hello " (world) "!"
}

// after (no recompilation needed, as only static content changed)
html! {
    "Goodbye " (world) "!"
}
```

but this change requires proper recompilation:

```rust
// before
html! {
    "Hello " (world) "!"
}

// after (recompilation needed, as rust code changed)
html! {
    "Hello " (other_world) "!"
}
```

the compiled code is very inefficient, as it re-parses rust sourcecode at runtime (still faster than rustc). there are some complications, such as the hot-reloading:

* not working in doctests
* not working whenever the `html!` macro is aliased or used from within another macro
* not working when `html! { }` is written in a weird way, such as `html! /* { */ {`

## Is this really for maud?

Even though there are no added dependencies, the feature is still a lot to maintain and required a lot of restructuring. I think most of the restructuring was unavoidable, and now there are three crates instead of two.

I don't necessarily have expectations that this will be merged, but on the other hand I also really want a template engine that is type-safe, has fast iteration-speeds and can be composed in a JSX-like way using regular Rust functions. I think maud+hotreload is the only standalone templating engine in the Rust ecosystem that can do that right now. So I'd be willing to fork maud and continue development there if people are interested. It seems there are a few other PRs that need attention anyway.

finally, thanks to @wrapperup for the initial prototype and architecture of hot-reloading.


jrey8343:
This is exactly the feature I’m looking for! 
Following 🍿🍿

schalk-b:
This is a great contribution! Exactly what I'm looking for. I'm in favour of a fork with new contributors if this isn't being actively maintained.

emschwartz:
I would also love something like this. I really like the experience of using Maud but the reload time makes web development quite a bit clunkier than working with a JS framework.

PacificBird:
If this doesn't get merged I'd love to see a fork with this functionality! I am currently using a super simple Poem-Maud-HTMX stack and this would improve the dev experience dramatically!

tryoxiss:
When compiling a release build it should always be static-only, no need to reconstruct trees at runtime for compile speed when compile speed isn't a concern :D

eboody:
I might be misunderstanding the problem-domain here but how is this different than using [tower-reload](https://github.com/leotaku/tower-livereload) and running `watchexec --restart --exts rs,html,css,js -- cargo run`?

PR #456, opened 2024-12-30
Introducing feature 'alloc' that removes dependency of alloc::sync::Arc

jakobdalsgaard:
The naming might be off here -- didn't really know a better feature name.

I'm using Maud for esp no_std development, and the esp-alloc crate I use there, does not have sync::Arc -- so I made a feature in Maud called 'alloc' (perhaps it should be called 'esp-alloc'?) that when enabled, will have Maud skip support for sync::Arc. Works great so far :-)

### DIFF
--- a/maud/Cargo.toml
+++ b/maud/Cargo.toml
@@ -14,6 +14,7 @@ edition.workspace = true
 
 [features]
 default = []
+alloc = []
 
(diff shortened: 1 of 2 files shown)


lambda-fairy:
Closed in favor of #492, thanks!

PR #457, opened 2025-01-01
Bump axum-core to 0.5 for axum 0.8 compatibility

### LINKED ISSUE #453: Axum integration broken again
Just a heads up that the 0.8 release of Axum (which is currently in alpha) will break the maud integration again. The error is exactly the same as described in this prior issue (and several others):

https://github.com/lambda-fairy/maud/issues/414

masklinn:
axum 0.8 uses axum-core 0.5, which is not semver-compatible with 0.4 and thus doesn't get picked up.

Not sure if there's more to do.

Fixes #453

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- Support `axum` v0.8 through `axum-core` v0.5
+
 ## [0.26.0] - 2024-01-15
 
 - Remove `AsRef<str>` restriction from `PreEscaped`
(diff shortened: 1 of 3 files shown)


lambda-fairy:
Thank you!

carlos-verdes:
on which version is this fix included?

masklinn:
I don't think a release has been cut yet, and LF seems to be going through the PR backlog a bit while they're available / have the time. If you need axum 0.8 urgently you may want to temporarily switch to a direct git dependency temporarily.

PR #412, opened 2024-01-07
feat: modernize parser

vidhanio:
parser rewritten from the ground-up using `syn` idioms and more strongly-typed `syn` structures rather than `TokenStream`s. passes all test cases flawlessly, and makes the code much easier to understand.

also drops `proc-macro-error` for a more robust solution (`proc-macro2-diagnostics`) which doesn't need global state and macros.

all of the edited trybuild test-cases are either adding more information/cases or fixing some bugs.

PR #459, opened 2025-01-07
Disallow elements in attributes

lambda-fairy:
Fixes a regression from #412.

The following code was allowed, and this PR fixes the parser to ban it:

```rust
html! {
    a href={ b {} } {}
}
```


lambda-fairy:
I wonder if a type parameter approach could be cleaner, like: https://github.com/vidhanio/hypertext/blob/caaff0a3f2250485af7a9acf565a6e470178cca9/hypertext-macros/src/maud.rs#L171

PR #466, opened 2025-01-15
feat: Add support for web framework poem@3

glennib:
Support is gated behind the poem-3 feature.

### DIFF
--- a/maud/Cargo.toml
+++ b/maud/Cargo.toml
@@ -18,6 +18,7 @@ default = []
 # Web framework integrations
 actix-web = ["actix-web-dep", "futures-util"]
 axum = ["axum-core", "http"]
+poem-3 = ["poem"]
 
 [dependencies]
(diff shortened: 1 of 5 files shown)


glennib:
@lambda-fairy do you think you can look at this PR?

maud/Cargo.toml
```diff
@@ -18,6 +18,7 @@ default = []
 # Web framework integrations
 actix-web = ["actix-web-dep", "futures-util"]
 axum = ["axum-core", "http"]
+poem-3 = ["poem"]
```
lambda-fairy, line 21: Can we remove this and name the feature `poem` instead (without the number)?

We only support the latest version of every framework, so the `3` is redundant.

glennib: Sure! Thanks for taking a look.

lambda-fairy:
Thanks!

PR #467, opened 2025-01-20
Avoid allocation when converting to axum's Response

dpc:
I have not bother benchmarking, but I believe avoiding allocating a HashMap will be easier to optimize and/or faster at runtime.

### DIFF
--- a/maud/src/lib.rs
+++ b/maud/src/lib.rs
@@ -368,11 +368,10 @@ mod axum_support {
 
     impl IntoResponse for PreEscaped<String> {
         fn into_response(self) -> Response {
-            let mut headers = HeaderMap::new();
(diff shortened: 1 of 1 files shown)


dpc:
Wait a moment. The `.0` there is pre-escaped, no? Doesn't this require some ... rendering step? I'm probably missing something, but just wondering. BTw. Thanks for `maud`, I really enjoy using it.

nubunto (approved):
LGTM!

lambda-fairy:
Thanks!

> The `.0` there is pre-escaped, no? Doesn't this require some ... rendering step?

Yes, the `html!` outputs a `String` which is packed into the `PreEscaped`.

PR #475, opened 2025-04-03
Ensure that the readme is displayed correctly on crates.io

Kijewski:
* Explicitly include readme and license files, so that they are available in crates.io and docs.rs.
* Use absolute URL to logo in readme, so that the logo is displayed correctly on crates.io.


lambda-fairy:
Thanks for the PR.

For `license-file`, can we [set it in the workspace](https://doc.rust-lang.org/cargo/reference/workspaces.html#the-package-table) instead?

For `readme`, I believe setting `readme = "../README.md"` works.

I'd like to avoid using softlinks as they aren't supported on Windows.

lambda-fairy:
Sounds reasonable. Thanks!

PR #484, opened 2025-11-11
Add macro which uses pre-existing buffer for rendering

### LINKED ISSUE #381: Make nested `html!` invocations lazy
Related to #373.

For patterns like this:

```rust
fn body() -> Markup { ... }

html! {
  h1 { "My page" }
  main { (body()) }
}
```

There is technically no need for the intermediate buffer from `body`, because the markup could render directly into the outer buffer.

In my ideal API, `Markup` would just implement `Render` (the `render` method would have to return something other than `Markup`, probably just `PreEscaped<String>`) and take advantage of the existing `render_to` method.

StratusFearMe21:
This PR adds the `html_render` macro to this crate which produces a struct that is able to both render to a pre-existing buffer, *and also* be passed into another invocation of `html` and render inside of that without allocating a new buffer.

Here's the documentation I created for the new `html_render` macro:

Functionally the same as [`html`] but produces a struct that implements [`Render`].

You can

- Call `render()` to render the HTML as [`Markup`]
- Call `render_to()` to render the HTML to a pre-existing `&mut String`
- Pass this directly into an [`html`] invocation to add it as a component without allocating a new buffer.

```rust
use maud::{html, html_render, DOCTYPE};

let component = html_render! {
   p { "Wow, such component!" }
};

let full = html! {
   (DOCTYPE)
   head {}
   html {
       (component)
   }
};
```

This PR resolves #381 and #90

### DIFF
--- a/maud/benches/complicated_maud.rs
+++ b/maud/benches/complicated_maud.rs
@@ -2,7 +2,7 @@
 
 extern crate test;
 
-use maud::{html, Markup};
+use maud::{html, html_render, Markup, Render};
 
 #[derive(Debug)]
 struct Entry {
@@ -11,7 +11,7 @@ struct Entry {
 }
 
 mod btn {
-    use maud::{html, Markup, Render};
+    use maud::{html_render, Render};
 
     #[derive(Copy, Clone)]
     pub enum RequestMethod {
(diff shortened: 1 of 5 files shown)


StratusFearMe21:
The performance wins are indeed significant

image

PR #495, opened 2026-04-06
Ensure no_std builds

AverageHelper:
Follow-up to #492, this PR adds a CI step to ensure maud continues to build in `no_std` environments (in this case, the `thumbv6m-none-eabi` target, which has only `core` and `alloc`, minus atomic pointers).


lambda-fairy:
Thanks!

Issue #498, opened 2026-05-01
Debian: Failing tests


kpcyrd:
hello! I'm trying to package the maud 0.27.0 crate into Debian, I noticed the following tests though - I'm not sure why this breaks, and it might be due to the versions of the other crates we have in Debian, but I was wondering if you have any hints/suggestions since you're more familiar with the codebase? :) Thank you!

```
test tests/warnings/attribute-missing-value.rs ... ok
test tests/warnings/class-shorthand-missing-value.rs ... ok
test tests/warnings/dynamic-attribute-names.rs ... ok
test tests/warnings/elements-in-attributes.rs ... ok
test tests/warnings/keyword-without-at.rs ... mismatch

EXPECTED:
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈
error: found keyword `if`
 --> tests/warnings/keyword-without-at.rs:5:9
  |
5 |         if {}
  |         ^^
  |
  = help: should this be `@if`?

error: found keyword `else`
 --> tests/warnings/keyword-without-at.rs:6:9
  |
6 |         else {}
  |         ^^^^
  |
  = help: should this be `@else`?

error: found keyword `for`
 --> tests/warnings/keyword-without-at.rs:7:9
  |
7 |         for {}
  |         ^^^
  |
  = help: should this be `@for`?

error: found keyword `while`
 --> tests/warnings/keyword-without-at.rs:8:9
  |
8 |         while {}
  |         ^^^^^
  |
  = help: should this be `@while`?

error: found keyword `match`
 --> tests/warnings/keyword-without-at.rs:9:9
  |
9 |         match {}
  |         ^^^^^
  |
  = help: should this be `@match`?
┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈┈

ACTUAL OUTPUT:
[...]

PR #500, opened 2026-05-11
Handle stable warning snapshots

tdashelby-cmyk:
This addresses #498.

The warning tests compare trybuild stderr fixtures. The existing fixtures match the nightly diagnostic rendering used by the repository CI, while the Debian report shows the stable-style rendering for three diagnostics where help notes are printed before the source location.

This keeps the current nightly warning run unchanged and adds stable-specific fixtures for those three cases. On nightly, the test still runs `tests/warnings/*.rs`; on non-nightly, it runs the unaffected warning cases plus the stable fixtures.

Verification:
- `git diff --check`
- `git show --check --oneline HEAD`
- Compared the stable fixtures against the actual output in #498

I could not run `cargo test -p maud --test errors run_warnings` locally in this environment because `cargo` is not installed on PATH. The first CI run showed the original fixture replacement was wrong for nightly, so I amended this to keep the nightly fixtures unchanged and add non-nightly fixtures instead.

### DIFF
--- a/maud/Cargo.toml
+++ b/maud/Cargo.toml
@@ -43,6 +43,7 @@ serde_json = { version = "1", optional = true }
 
 [dev-dependencies]
 serde = { version = "1", features = ["derive"] }
+rustversion = "1"
 trybuild = { version = "1.0.33", features = ["diff"] }
 
 [package.metadata.docs.rs]
--- a/maud/tests/errors.rs
+++ b/maud/tests/errors.rs
@@ -1,7 +1,22 @@
 use trybuild::TestCases;
 
 #[test]
+#[rustversion::nightly]
 fn run_warnings() {
     let config = TestCases::new();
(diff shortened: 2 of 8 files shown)


tdashelby-cmyk:
CI reran after the branch update. The warning-test job (`Main`) is now passing, along with `Documentation` and `Rustfmt`.

The remaining red job is `Doctest`; it fails while compiling `rustix 0.37.28` with `cannot find attribute rustc_layout_scalar_valid_range_*`, before reaching this change's warning tests.

lambda-fairy:
Thanks!

Rather than adding a separate version for stable, how about we disable the trybuild tests on non-nightly altogether?

The differences are minor and the impact of bad diagnostics is low. On the other hand, having two versions means that contributors need to do two separate builds before making any change to warning messages.

Issue #474, opened 2025-04-02
Script without body generates wrong html


niklaswimmer:
Hey, I tried to include `script src="main.js;` in my template's `head` which expands to `<script src="main.js">`. This is however wrong as script tags always need to be closed. The solution is to use `script src="main.js" {}` in the template instead. Maybe this can be improved by special casing the generation of `script` tags?

tryoxiss:
Not a maintainer, just trying to be helpful.

Maud does this for all elements, if it is ended with a semicolon it will not place an end tag as it assumes self-closing. If it is ended with braces it will place an end tag. `script src="main.js {}` will work as expected.

Though since HTML void elements are very limited -- you cant define custom void elements. It would make sense for maud to compile error (or at least raise a warning) when trying to use a void element as a block or a block element as a void element because all void elements must not have a closing tag and all non-void elements must have a closing tag (to my knowlege at least).

Complete list of void elements:

- `<area>`
- `<base>`
- `<br>`
- `<col>`
- `<embed>`
- `<hr>`
- `<img>`
- `<input>`
- `<link>`
- `<meta>`
- `<param>` (DEPRECATED)
- `<source>`
- `<track>`
- `<wbr>`

(Source: https://developer.mozilla.org/en-US/docs/Glossary/Void_element )

In other words, if any of these are terminated with brackets `{}`, it would error. If any other element is terminated with a semicolon `;`, it would error. If this were done, the `;` synax could even be completely removed since it would then be able to figure out how to end the element based on what it is.

> Sidenote: While we are on the topic of validating HTML, perhaps it coul also be a good idea to require elements to be valid aswell. Because all custom elements need to have a hyphen (`-`). Maud could reject any elements without a hyphen that arent in a list of valid elements.
> 
> Weather or not deprecated elements would be allowed or not im not sure, maybe it could emit a warning or bar behind a feature or something). It would be relatively simple to do this, since the rules for this are intentionally simple.
> 
> Validating attributes would also be possible but much harder. Attributes have much more complex rules, theres a lot more of them, and more are added much more frequently. While it would be nice to validate attributes, I dont suspect it would be worth it for the added complexity.

encounter:
How about if an element is void, `;` retains the current behavior: `img;` -> `<img>`. If an element is _not_ void, it emits a self-closing tag: `script;` -> `<script/>`.

[...]

Issue #432, opened 2024-07-01
empty tags using semicolons should create closing tags for non-void elements


twynb:
Take this example code:

```rust
html! {
    footer {
        div class="nav-padding";
        div class="nav-item" { "example1" }
        div class="nav-item" { "example2" }
        div class="nav-item" { "example3" }
        div class="nav-padding";
    }
}
```

I'd expect the `div`s closed by semicolons to become empty divs:

```html
<footer>
    <div class="nav-padding"></div>
    <div class="nav-item">example1</div>
    <div class="nav-item">example2</div>
    <div class="nav-item">example3</div>
    <div class="nav-padding"></div>
</footer>
```

The actual output is (with whitespace added for readability):

```html
<footer>
    <div class="nav-padding">
        <div class="nav-item">example1</div>
        <div class="nav-item">example2</div>
        <div class="nav-item">example3</div>
        <div class="nav-padding">
</footer>
```

This matches the behaviour documented in [Elements and attributes](https://maud.lambda.xyz/elements-attributes.html), only creating an opening tag (because that works for `<br>` tags), but is a needless pitfall.

My suggestion would be to either insert a closing tag for tags that need one (like `<div>`) or to have this throw an error within the macro - since you can't manually close a div tag opened using `div;`, this always leads to invalid/broken HTML.

LnxgpI0k:
I've encountered a bug: Generated singleton tags don't actually self-close, ie every one of them is missing a `/` before the alligator mouth. Should I open a separate issue?

RoyaltyInTraining:
Implementing either of the solutions would be a massive improvement and save a lot of debugging time. I keep running into silent errors whenever I wrongfully assume that some element can be self-closed. I wish Maud would use the power of Rust's macro system in order to help users ensure that their HTML is valid.

PR #505, opened 2026-08-02
Add check for correct closing method of html element

RoyaltyInTraining:
This addresses problems which multiple issues have mentioned (#432, #474).

The problem is: Block elements closed with `;` and void elements closed with `{}` currently generate invalid HTML. Especially the block elements which are missing their closing tags can easily cause unexpected results after the browser attempts to silently fix the error.

This PR adds simple validation logic which emits errors when a mismatch is detected. It uses the current list of [void elements in the HTML living standatd](https://html.spec.whatwg.org/multipage/syntax.html#elements-2). The generation logic wasn't changed, so the output of valid usage remains identical. A `html_unchecked` macro is also included as an escape hatch.


LnxgpI0k:
Does this enable optionally closing regular tags like void elements?

RoyaltyInTraining:
> Does this enable optionally closing regular tags like void elements?

No, it only helps developers to write correct code by highlighting mistakes. I know that some people want the added syntax, but since this is my first contribution to this project, I chose to not touch the `ast` module. I'm also not sure if the repo owner even wants this additional syntax rule, or if they prefer the current 1:1 mapping to html.

